PR #11, opened 2019-09-19
Docs: Overview (Concepts and Architecture)

jameshochadel:


docs/README.md
```diff
@@ -2,6 +2,7 @@
 
 - [Introduction](introduction.md)
 - [Quick Start](quickstart.md)
+- [Overview](overview.md)
```
jgustie, line 5: Is it too much to have "Introduction" and "Overview"? It feels like ski trail names: "Upper Introduction", "Lower Introduction", "Middle Overview", "Lower-mid Introduction"...

jgustie (approved):

docs/README.md
jameshochadel: Perhaps. My original name for it was Concepts, but the section itself broke naturally into Architecture and Concepts, so that didn't seem right. Perhaps I change this to `concepts.md` and replace the Concepts header in the doc with... "Components"? "Experiments," since that's the top-level concept?

jameshochadel: @jgustie Should I revise this or merge as-is?

jgustie: It's fine as is for now.

PR #15, opened 2019-10-01
Revise Quickstart to be testable with mdrip

jameshochadel:
@jgustie This isn't ready to merge yet, but I wanted to get your feedback early, particularly on the way I'm using `expect`. Does this look alright to you?

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -20,7 +20,7 @@ For more details, see [the installation guide](install.md).
 
 Once you have the Red Sky Ops Tool you can initialize the manager in your cluster:
 
-<!-- @init -->
+<!-- @init @test -->
 ```sh
 redskyctl init
 ```
@@ -29,7 +29,7 @@ redskyctl init
 
 Generally you will want to write your own experiments to run trials on your own applications. For the purposes of this guide we can use the simple example found in the `k8s-experiment` [repository on GitHub](https://github.com/redskyops/k8s-experiment/tree/master/examples/simple):
 
-<!-- @apply -->
+<!-- @apply @test -->
 ```sh
 kustomize build github.com/redskyops/k8s-experiment//examples/simple | kubectl apply -f -
 ```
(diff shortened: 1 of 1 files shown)


jgustie:
You don't need to use `expect`, `redskyctl suggest` doesn't need to run interactively, it will take [assignments as arguments](https://redskyops.dev/docs/redskyctl/#redskyctl-suggest), e.g. `redskyctl suggest simple -A batchSize=256 --assign workers=2`; you can even just do a `--default rand` and get random suggestions.

jameshochadel:
Nice!! I will do that.

docs/quickstart.md
```diff
@@ -20,32 +20,46 @@ For more details, see [the installation guide](install.md).
...
+> ```
+
+You will be prompted to enter a value for each parameter in the experiment and a new trial will be created.
+
+You can monitor the progress using `kubectl`:
 
+<!-- @getTrials -->
 ```sh
-$ kubectl get trials
+kubectl get trials
```
robpacheco: What validates that this ran successfully?

jameshochadel: Good question.`mdrip` defines success as "every command exited with code 0". So this block will not validate anything other than that `kubectl get trials` runs without error. How about I change it to:

```
kubectl get trials | grep simple
```

...which will fail if nothing in the list matches?

jameshochadel: (There are alternative ways we could achieve that of course, but I think the `grep` technique strikes the right balance between making the test better and keeping the docs readable.)

jgustie: Why not `kubectl get trial simple`?

jameshochadel: Because `kubectl get` looks for an exact match, so that'll fail due to the suffix on the trial.

jgustie: Good point, how about `redskyctl get trials simple` then?

PR #17, opened 2019-10-11
added first version of resource query

jdplatt:
Here is a first version of cost query. I tested it with the following query and it all worked great.

 - name: cost
    minimize: true
    type: pods
    query: '{{resource .Pods "cpu=22,memory=3"}}'
    selector:
      matchLabels:
        app: voting-app

Let me know if you want me to change the format for the weights or the unit conversions I did to make the weights easy to understand (memory is /GB and cpu is /core).

### DIFF
--- a/pkg/controller/template/function.go
+++ b/pkg/controller/template/function.go
@@ -34,9 +34,9 @@ func FuncMap() template.FuncMap {
 	delete(f, "expandenv")
 
 	extra := template.FuncMap{
-		"duration":  duration,
-		"percent":   percent,
-		"resources": resources,
+		"duration":         duration,
+		"percent":          percent,
+		"resourceRequests": resourceRequests,
 	}
 
 	for k, v := range extra {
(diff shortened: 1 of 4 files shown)


pkg/controller/template/function.go
```diff
@@ -19,10 +19,13 @@ package template
 import (
 	"fmt"
 	"reflect"
+	"strconv"
+	"strings"
 	"text/template"
 	"time"
 
 	"github.com/Masterminds/sprig"
+	v1 "k8s.io/api/core/v1"
```
jgustie: Change this import to `corev1 "k8s.io/api/core/v1"` for consistency.

pkg/controller/template/function.go
```diff
@@ -37,6 +40,7 @@ func FuncMap() template.FuncMap {
 		"duration": duration,
 		"percent":  percent,
 		"sum":      sum,
+		"resource": totalResources,
```
jgustie: Keep the names consistent in the code and the template engine.

jgustie (approved):
Name of this function will probably change to `resourceRequests` prior to release of 1.2.1 though...

...ah...you already beat me to it...

PR #32, opened 2020-04-02
Add tests for internal/controller

andrewhare:
This adds tests for `internal/controller`.

```
✝ go test -cover -race ./internal/controller
ok      github.com/redskyops/redskyops-controller/internal/controller   0.724s  coverage: 95.1% of statements
```

@jgustie This required a little bit of refactoring to get test coverage over 90%

### DIFF
--- a/internal/controller/result.go
+++ b/internal/controller/result.go
@@ -17,6 +17,10 @@ limitations under the License.
 package controller
 
 import (
+	"path"
+	"runtime"
+	"strings"
+
 	redskyapi "github.com/redskyops/redskyops-controller/redskyapi/experiments/v1alpha1"
 	apierrs "k8s.io/apimachinery/pkg/api/errors"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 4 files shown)


internal/controller/result.go
```diff
@@ -38,11 +42,43 @@ func RequeueIfUnavailable(err error) (*ctrl.Result, error) {
...
-		ReconcileConflictErrors.WithLabelValues(guessController()).Inc()
+		controllerName := guessController(fs)
+		ReconcileConflictErrors.WithLabelValues(controllerName).Inc()
 		result.Requeue = true
 		err = nil
 	}
 	return result, err
 }
+
+// guessController dumps stack to try and guess what the controller name should be
```
jgustie, line 59: This was a hack when I wanted to get metrics in so we could keep track of where we were seeing conflicts. In hindsight it would probably be better if we did something similar to what we do with logs; that is have a `Metrics` interface with a method like `ReconcileConflictErrors() prometheus.Counter` and store and instance with the populated labels in a field on relevant `*Reconciler` types (populated in `main.go`). We could hang the labelled metrics instance off the context at the start of the `Reconcile` method and change the signature above to `RequeueConflict(ctx context.Context, err error)...`.

jgustie (approved):

PR #33, opened 2020-04-06
GitHub Actions

jgustie:
This switches from strictly Make/Docker + CircleCI to Make/Docker/GoReleaser + GitHub Actions.

I am starting with two separate "workflows", one for changes to the master branch and one for pull requests to the master branch. Both workflows will push images to GCR (since using built images out of the GCR is the easiest way to get pre-release builds into the hands of the testers).

For the most part, things have been ported 1-1 from CircleCI to GitHub Actions; with the exception of the switch to GoReleaser for `redskyctl` (which eliminated the need for `ghr` and will eventually lead to other build improvements like signing). Outside of CI, the `make tool` target will mimic the snapshot release for local builds. Also, the Docker image now uses the already build binary instead of rebuilding it.

Finally, the versioning scheme has changed. Previously the `version.go` file was updated at or after each release to contain the "next" version to be released. That is no longer the case and the alternate behavior is to override the variables in `version.go` on every build: if a version is not explicitly set, the Makefile will set it to the tag of the last release plus `-next` (e.g. at the time of writing we will go from `1.5.0` to `1.4.2-next`).


jgustie:
Failing checks are due to removed configurations.

PR #34, opened 2020-04-13
Add parameter constraints

jgustie:


internal/server/server.go
```diff
@@ -68,6 +68,43 @@ func FromCluster(in *redskyv1alpha1.Experiment) (redskyapi.ExperimentName, *reds
 		})
 	}
 
+	out.Constraints = nil
```
andrewhare, line 71: nitpick: Initialize the capacity of `out.Constraints`:

```go
out.Constraints = make([]redskyapi.Constraint, 0, len(in.Spec.Constraints))
```

You can do the same thing on line 108 for metrics as well.

internal/server/server.go
```diff
@@ -68,6 +68,43 @@ func FromCluster(in *redskyv1alpha1.Experiment) (redskyapi.ExperimentName, *reds
...
+				ConstraintType: redskyapi.ConstraintOrder,
+				OrderConstraint: redskyapi.OrderConstraint{
+					LowerParameter: c.Order.LowerParameter,
+					UpperParameter: c.Order.UpperParameter,
+				},
+			})
+		case c.Sum != nil:
+			sc := redskyapi.SumConstraint{
+				IsUpperBound: c.Sum.IsUpperBound,
+				Bound:        float64(c.Sum.Bound.MilliValue()) / 1000,
```
andrewhare, line 86: From reading the docs and source of [`resource.Quantity`](https://github.com/kubernetes/apimachinery/blob/06deae5c9c2c030d771a467e086b6c791e8800dc/pkg/api/resource/quantity.go#L695) it seems like we should be able to do this and avoid any math and/or potential overflowing of an `int64`:

```go
Bound: float64(c.Sum.Bound.Value())
```

Same thing on line 96.

andrewhare (changes requested):
A couple of nitpicks but otherwise LGTM.

internal/server/server.go
jgustie: Yeah. I'm an idiot: `x*1000/1000` is still `x`. No PhD in Maths required.

internal/server/server.go
jgustie: I had considered that (in several places), but ultimately the numbers are so small (e.g. this function is called once over the lifetime of an experiment on slices that contain only a few elements) that I opted for the simple code. In this particular case the `nil` slice also serializes differently then an empty slice (which impacts test and network representations).

internal/server/server.go
jgustie: ...except when it's not.

I always forget how quantity scaling works: the scaling is applied before the rounding so `MustParse("3.5").Value() == 4` while `MustParse("3.5").MilliValue() == 3500`.

I'd like to pretend the insight to test to with `3.5` was intentional, but alas, it was dumb luck.

andrewhare: Nice catch, that's a good point about the rounding.

andrewhare (approved):

PR #38, opened 2020-04-20
Clean up some error messages

bradbeam:
- Mark filename flag required for `redskyctl generate trial`
- Return an error message when supplied value is out of bounds for the experiment


jgustie (approved):

bradbeam:
forks are no bueno

PR #40, opened 2020-04-20
fix: Generate random suffix for trial

bradbeam:
Fixes issue with generated name ending in `-`.
Fixes missing trial.Name so it is a valid resource.

### DIFF
--- a/internal/experiment/trial_template.go
+++ b/internal/experiment/trial_template.go
@@ -17,6 +17,9 @@ limitations under the License.
 package experiment
 
 import (
+	"fmt"
+	"math/rand"
+
 	redskyv1alpha1 "github.com/redskyops/redskyops-controller/pkg/apis/redsky/v1alpha1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


jgustie:
This breaks the code that inserts the trial number from server (in place of the random suffix you are supposed to get from using `generateName`), is just specifying the `generateName` for the trial not working for some/all versions of the Kube API?

jgustie:
It looks like I might have [anticipated this](https://github.com/redskyops/redskyops-controller/blob/fd6beb659e22a2a45778d562cbdc9638baf889a8/redskyctl/internal/commands/generate/trial.go#L87). Instead of doing this as written, put the random generation where that TODO is (similar to what we do when using the [remote server](https://github.com/redskyops/redskyops-controller/blob/fd6beb659e22a2a45778d562cbdc9638baf889a8/internal/server/server.go#L145), except using a random suffix in place of the number scavenged from the URL)

bradbeam:
Closing as not needed. 
When using `generateName`, `kubectl apply` doesn't work, need to use `kubectl create` instead.

PR #45, opened 2020-04-24
test: Add tests for metric controller

bradbeam:
- Added some tests around collectMetrics
- Updated github workflow to install kubebuilder so we have api server and etcd available for the testenvironment

### DIFF
--- a/.github/workflows/master.yaml
+++ b/.github/workflows/master.yaml
@@ -46,6 +46,13 @@ jobs:
           echo "::set-env name=PULL_POLICY::Always"
           echo "::set-env name=DOCKER_TAG::${DOCKER_TAG}"
           gcloud --quiet auth configure-docker
+
+          # kubebuilder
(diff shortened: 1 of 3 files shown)


andrewhare:
This looks cool. Is `sigs.k8s.io/controller-runtime/pkg/envtest` like `net/http/httptest` or does it spin up actual k8s infrastructure?

I'm wondering if this test still qualifies as a unit test or if it's more of an integration test.

bradbeam:
It spins up an API Server and etcd instance. It's a bit limited in that it won't actually schedule anything ( no nodes, scheduler, controller manager ), but enough to play with api objects and let the clients act against the api.

jgustie:
This un-does the change from 273950218cf0caba113d573688d65dc10daf7a15 (explicitly removing the controller runtime envtests). Let's take it offline on Monday.

PR #49, opened 2020-04-28
Silence stderr on the forked exec

jgustie:
I want to get this in before the release because it doesn't look good, but it's not a long term "solution". I believe there are other intermittent failures occurring here as well: but it needs more investigation.

### DIFF
--- a/redskyctl/internal/commands/initialize/generator.go
+++ b/redskyctl/internal/commands/initialize/generator.go
@@ -85,12 +85,13 @@ func (o *GeneratorOptions) generate(ctx context.Context) error {
 	args = append(args, "--", "install")
 
 	// Run the command straight through to the configured output stream
-	// TODO How do we filter out the warning about not being able to attach?
 	kubectlRun, err := o.Config.Kubectl(ctx, args...)
 	if err != nil {
 		return err
 	}
+	// TODO Should we buffer this and verify we got valid YAML?
 	kubectlRun.Stdout = o.Out
-	kubectlRun.Stderr = o.ErrOut
+	// TODO What is the best way to filter out the "Error attaching, ..." line from the writer?
+	// kubectlRun.Stderr = o.ErrOut
 	return kubectlRun.Run()
 }


andrewhare:
Is there somewhere else we can pipe stderr (logfile, etc)? Silencing it altogether seems problematic even in the short term.

redskyctl/internal/commands/initialize/generator.go
```diff
@@ -85,12 +85,13 @@ func (o *GeneratorOptions) generate(ctx context.Context) error {
...
-	// TODO How do we filter out the warning about not being able to attach?
 	kubectlRun, err := o.Config.Kubectl(ctx, args...)
 	if err != nil {
 		return err
 	}
+	// TODO Should we buffer this and verify we got valid YAML?
 	kubectlRun.Stdout = o.Out
-	kubectlRun.Stderr = o.ErrOut
+	// TODO What is the best way to filter out the "Error attaching, ..." line from the writer?
+	// kubectlRun.Stderr = o.ErrOut
```
andrewhare, line 95: Use a [`bufio.Scanner`](https://golang.org/pkg/bufio/#Scanner) to filter the stream.

bradbeam:
Not sure I'm familiar with this, have sample output for context?

jgustie:
The initialization generator (currently) works by doing a `kubectl run` on the `setuptools` image to generate all the manifests for installation. Occasionally the exec pod finishes before `kubectl` can attach to the container's output; if this happens `kubectl` falls back to getting the pod logs (which in this case contain pure YAML). It spits out a message on stderr indicating this is what happened: but that message has no impact on the success or failure of the generation of the YAML and just confuses people.

I also suspect the exec fails occasionally (perhaps due to a timeout) and the result is error text ends up on stdout. In this case the generator should fail immediately instead of waiting for `kubectl` to fail when it tries to parse the error messages as YAML.

jgustie:
Brad, I'm glad you haven't seen it yet. I had to do a couple of init/resets before I got the output:
```
➜ redskyctl init 
Error attaching, falling back to logs: 
namespace/redsky-system created
customresourcedefinition.apiextensions.k8s.io/experiments.redskyops.dev created
customresourcedefinition.apiextensions.k8s.io/trials.redskyops.dev created
clusterrole.rbac.authorization.k8s.io/redsky-manager-role created
clusterrole.rbac.authorization.k8s.io/redsky-patching-role created
clusterrole.rbac.authorization.k8s.io/redsky-proxy-role created
clusterrole.rbac.authorization.k8s.io/redsky-metrics-reader created
clusterrolebinding.rbac.authorization.k8s.io/redsky-manager-rolebinding created
clusterrolebinding.rbac.authorization.k8s.io/redsky-patching-rolebinding created
clusterrolebinding.rbac.authorization.k8s.io/redsky-proxy-rolebinding created
service/redsky-controller-manager-metrics-service created
deployment.apps/redsky-controller-manager created
clusterrole.rbac.authorization.k8s.io/redsky-aggregate-to-patching created
secret/redsky-manager created
```

bradbeam:
>  if this happens kubectl falls back to getting the pod logs

Given the example output above and this statement, is there any advantage to capturing the stdout/err instead of just pulling the container log? You mentioned that the logs would contain yaml, but I don't see that above.

jgustie:
Basically what you are seeing in that example (i.e. what `redskyctl init` does) is:
```
kubectl run ... --image setuptools| kubectl apply -f -
```
The "Error attaching..." comes from the stderr of the `run` and everything else comes from the stdout of `apply`.

redskyctl/internal/commands/initialize/generator.go
bradbeam: Think this seems like the best option right now

jgustie:
Ugh. I just saw this gem (all one line):
```
Error attaching, falling back to logs: unable to upgrade connection: container redsky-bootstrap not found in pod redsky-bootstrap_default
```

bradbeam (approved):

PR #53, opened 2020-04-30
chore: Update license header year

bradbeam:
Also ensure consistent license header format for all files.


andrewhare (approved):

andrewhare:
We should add a pre-commit hook to do this automatically from now on.

bradbeam:
Added a step in the gh actions workflow to check/catch this

PR #57, opened 2020-05-01
Add initial delay seconds to the trial

jgustie:


jdplatt (commented):
Is this the same code we had before?

jgustie:
Yes, minus the check to see if `trial-ready == False` (which can't happen).

jdplatt (approved):

bradbeam (approved):

PR #60, opened 2020-05-04
Allow authorize cluster to generate Helm values

jgustie:
Just realized that there is no easy way for "enterprise customers" to install via Helm. This adds a `--helm-values` option to the `redskyctl authorize-cluster` command so instead of updating the secret directly in the cluster, it will print out a Helm `values.yaml` file for our chart.

Note that nothing prevents customers from running the regular `authorize-cluster` after a Helm install, but this might fit what they are expecting a little better.

### DIFF
--- a/redskyctl/internal/commands/authorize_cluster/generator.go
+++ b/redskyctl/internal/commands/authorize_cluster/generator.go
@@ -197,25 +197,25 @@ type helmValuesPrinter struct {
 }
 
 func (h helmValuesPrinter) PrintObj(i interface{}, w io.Writer) error {
-	if secret, ok := i.(*corev1.Secret); ok {
-		vals := map[string]interface{}{
-			"remoteServer": map[string]interface{}{
-				"enabled":      true,
-				"identifier":   string(secret.Data["REDSKY_SERVER_IDENTIFIER"]),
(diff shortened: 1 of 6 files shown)


redskyctl/internal/commands/authorize_cluster/generator.go
```diff
@@ -182,3 +192,30 @@ func (o *GeneratorOptions) clientInfo(ctx context.Context, ctrl *config.Controll
 	}
 	return o.Config.RegisterClient(ctx, client)
 }
+
+type helmValuesPrinter struct {
+}
+
+func (h helmValuesPrinter) PrintObj(i interface{}, w io.Writer) error {
+	if secret, ok := i.(*corev1.Secret); ok {
```
bradbeam: nit, Could we return early here?

```
secret, ok := i.(*corev1.Secret)
if !ok {
  return nil
}

vals ....
```

bradbeam (approved):

redskyctl/internal/commands/authorize_cluster/generator.go
jgustie: Yeah, I should probably be better about that in general...

PR #61, opened 2020-05-05
Show unique experiment name instead of display name

jgustie:
James first saw this as "experiment-not-found" when trying to get or delete an experiment. It turns out the problem was that we were showing the wrong name.

### DIFF
--- a/redskyapi/experiments/v1alpha1/api.go
+++ b/redskyapi/experiments/v1alpha1/api.go
@@ -274,6 +274,19 @@ type Experiment struct {
 	Parameters []Parameter `json:"parameters"`
 }
 
+// Name allows an experiment to be used as an ExperimentName
+func (e *Experiment) Name() string {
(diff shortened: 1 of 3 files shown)


redskyapi/experiments/v1alpha1/api.go
```diff
@@ -520,7 +533,7 @@ func (h *httpAPI) GetAllExperimentsByPage(ctx context.Context, u string) (Experi
 }
 
 func (h *httpAPI) GetExperimentByName(ctx context.Context, n ExperimentName) (Experiment, error) {
-	u := h.client.URL(endpointExperiment + url.PathEscape(n.Name()))
+	u := h.client.URL(endpointExperiment + n.Name())
```
jgustie, line 536: The escaping was not necessary here; at first I thought this was the bug (e.g. I was seeing `%2520` in the effective URL).

bradbeam (approved):

PR #63, opened 2020-05-05
Last call for v1alpha1

jgustie:
This is the last call for _minor_ changes to v1alpha1 in the 1.5.x series.

Basically I fixed the comments that are generating lint warnings so we are not eternally plagued by them once we "lock" v1alpha1.

Additionally, after a conversation with Steph, it appears that the some of the "required/optional" flags are a little off. I suspect things like parameters and metrics have always been "required" however they were marked as optional to get the Kubebuilder scaffolded tests to pass (those tests have since been removed altogether). Likewise, the "template" has always been optional, however we represent it as a struct (not a pointer to a struct) so it is always going to be there (thus the lack of an `omitempty` that would otherwise trigger the documentation to claim it is not necessary).

### DIFF
--- a/api/v1alpha1/experiment_types.go
+++ b/api/v1alpha1/experiment_types.go
@@ -77,33 +77,22 @@ type SumConstraint struct {
 	Parameters []SumConstraintParameter `json:"parameters"`
 }
 
-// PatchType represents the allowable types of patches
-type PatchType string
-
 // MetricType represents the allowable types of metrics
 type MetricType string
 
 const (
-	// Strategic merge patch
-	PatchStrategic PatchType = "strategic"
-	// Merge patch
-	PatchMerge = "merge"
-	// JSON patch (RFC 6902)
-	PatchJSON = "json"
-
-	// Local metrics are Go Templates evaluated against the trial itself. No external service is consulted, primarily
+	// MetricLocal metrics are Go Templates evaluated against the trial itself. No external service is consulted, primarily
 	// useful for extracting start and completion times.
 	MetricLocal MetricType = "local"
-	// Pod metrics are similar to local metrics, however the list of pods in the trial namespace matched by the selector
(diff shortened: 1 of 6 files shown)


bradbeam (approved):
Should we also use the most recent kubebuilder to generate the CRD manifests?
Locally I get changes to these files when they're generated:

```
diff --git a/config/crd/bases/redskyops.dev_experiments.yaml b/config/crd/bases/redskyops.dev_experiments.yaml
index 02ecd47b..6b58f66c 100644
--- a/config/crd/bases/redskyops.dev_experiments.yaml
+++ b/config/crd/bases/redskyops.dev_experiments.yaml
@@ -4,7 +4,7 @@ apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.2.2
+    controller-gen.kubebuilder.io/version: v0.2.5
   creationTimestamp: null
   name: experiments.redskyops.dev
 spec:
@@ -19,7 +19,7 @@ spec:
     listKind: ExperimentList
     plural: experiments
     singular: experiment
-  scope: ""
+  scope: Namespaced
   subresources: {}
   validation:
     openAPIV3Schema:
@@ -50,7 +50,11 @@ spec:
                   sum:
                     properties:
                       bound:
-                        type: string
+                        anyOf:
+                        - type: integer
+                        - type: string
+                        pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
+                        x-kubernetes-int-or-string: true
...
```

jgustie:
We can't use the latest versions of `controller-gen` due to issues with how it generates JSON schema for `IntOrString` types (the `x-kubernetes-int-or-string` bit causes some compatibility issues).

PR #67, opened 2020-05-07
feat: Add quickstart/postgres to gh actions

bradbeam:

### DIFF
--- a/.github/workflows/push.yaml
+++ b/.github/workflows/push.yaml
@@ -53,3 +53,38 @@ jobs:
         with:
           name: redskyctl_linux_amd64
           path: dist/redskyctl-linux-amd64.tar.gz
+      - name: KinD (Kubernetes in Docker) Action
+        uses: engineerd/setup-kind@v0.3.0
+      - name: redsky quickstart
+        run: |
+          #echo "Install redskyctl"
+          #curl -sLo - https://github.com/redskyops/redskyops-controller/releases/download/v1.5.0/redskyctl-linux-amd64.tar.gz | \
+          #  tar -zxf - && \
+          #  chmod 755 ./redskyctl
+          echo "Init redskyops"
+          dist/redskyctl_linux_amd64/redskyctl init
+          echo "Install kustomize"
+          curl -s "https://raw.githubusercontent.com/\
+          kubernetes-sigs/kustomize/master/hack/install_kustomize.sh"  | \
+            bash
+          echo "Patch postgres deployment"
+          ./kustomize build github.com/redskyops/redskyops-recipes/postgres > postgres.yaml
+          # Need to change strategy here since we're limited to 2cpu total
+          sed -i -e '37i\  strategy:\n    type: Recreate' postgres.yaml
+          echo "Create postgres experiment"
+          kubectl apply -f postgres.yaml
+          echo "Create new trial"
+          dist/redskyctl_linux_amd64/redskyctl generate trial \
+            --assign memory=500 \
+            --assign cpu=100 \
+            -f <(kubectl get experiment postgres-example-test -o yaml) | \
+              kubectl create -f -
+          kubectl get trial -o wide
+          # Change this back to a higher value when we can schedule the trial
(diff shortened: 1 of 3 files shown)


.github/workflows/push.yaml
```diff
@@ -53,3 +53,43 @@ jobs:
...
+      - name: KinD (Kubernetes in Docker) Action
+        uses: engineerd/setup-kind@v0.3.0
+      - name: commands
+        run: |
+          kubectl get nodes -o wide
+          kubectl get node kind-control-plane -o yaml
+          kubectl get pods -n kube-system
+      - name: redsky quickstart
+        run: |
+          echo "Install redskyctl"
```
bradbeam: We should probably skip this and use the artifact we just built

bradbeam:
This adds ~4-5min to the build but I think it gives a good sanity check that things work as expected.

.github/workflows/push.yaml
jgustie: I think you can use:

```
- uses: actions/download-artifact@v2
  with:
    name: redskyctl_linux_amd64
    path: ...
```

.github/workflows/push.yaml
```diff
@@ -53,3 +53,43 @@ jobs:
...
+          curl -sLo - https://github.com/redskyops/redskyops-controller/releases/download/v1.5.0/redskyctl-linux-amd64.tar.gz | \
+            tar -zxf - && \
+            chmod 755 ./redskyctl
```
jgustie: Was this necessary? The permissions should be preserved.

.github/workflows/push.yaml
```diff
@@ -53,3 +53,43 @@ jobs:
...
+          echo "Init redskyops"
+          ./redskyctl init
+          echo "Install kustomize"
+          curl -s "https://raw.githubusercontent.com/\
+          kubernetes-sigs/kustomize/master/hack/install_kustomize.sh"  | \
+            bash
```
jgustie: I wonder if the `kubectl -k` option is enough: we don't actually do much with these examples.

jgustie:
4-5 minutes on top of the 8 that we have is getting pretty steep. Does it make sense to put this in a separate workflow?

.github/workflows/push.yaml
bradbeam: No go with `kubectl -k` unfortunately. Looks like Kubectl bundles kustomize v2 and we need v3 to work with our kustomize files. :(

.github/workflows/push.yaml
bradbeam: Should be able to drop this and use the precompiled version ( under `dist` ).

.github/workflows/push.yaml
```diff
@@ -53,3 +58,32 @@ jobs:
...
+          echo "Init redskyops"
+          dist/redskyctl_linux_amd64/redskyctl init
+          echo "Install kustomize"
+          curl -s "https://raw.githubusercontent.com/\
+          kubernetes-sigs/kustomize/master/hack/install_kustomize.sh"  | \
+            bash
+          echo "Patch postgres deployment"
+          ./kustomize build github.com/redskyops/redskyops-recipes/postgres > postgres.yaml
+          # Need to change strategy here since we're limited to 2cpu total
+          sed -i -e '37i\  strategy:\n    type: Recreate' postgres.yaml
```
jgustie: My vote here would be to create `/test/quickstart/kustomization.yaml` file:

```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

resources:
- github.com/redskyops/redskyops-recipes/postgres

patchesStrategicMerge:
- |-
  apiVersion: apps/v1
  kind: Deployment
  metadata:
    name: postgres
  spec:
    strategy:
      type: Recreate
```
That way you can just `kustomize build test/quickstart  | kubectl apply -f -`.

I'm on the fence about adding a `/test/quickstart/run.sh` to do everything from here down.

jgustie (approved):

PR #68, opened 2020-05-07
fix: Ignore invalid git tags

bradbeam:
This filters the tags we retrieve from git. Specifically we're ignoring ^{}
because they don't play nice with version metadata.

Minor change to the checkout workflow - fetch tags only instead of
entire history. Saw the example in the readme and seemed useful.

### DIFF
--- a/.github/workflows/master.yaml
+++ b/.github/workflows/master.yaml
@@ -15,8 +15,8 @@ jobs:
     steps:
       - name: Check out code
         uses: actions/checkout@v2
-      - name: Unshallow
-        run: git fetch --prune --unshallow
+      - name: Fetch tags
+        run: git fetch --depth=1 origin '+refs/tags/*:refs/tags/*'
       - name: Set up Go
         uses: actions/setup-go@v2
         with:
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ GOBIN=$(shell go env GOBIN)
 endif
 
 # Collect version information
-VERSION ?= $(shell git ls-remote --tags origin 'v*' | tail -1 | awk -F/ '{ print $$3 }')-next
+VERSION ?= $(shell git ls-remote --tags --refs origin 'v*' | awk -F/ '{ print $$3 }' | tail -1)-next
 BUILD_METADATA ?=
 GIT_COMMIT ?= $(shell git rev-parse HEAD)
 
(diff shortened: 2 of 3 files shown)


jgustie:
Why not just do [this](https://github.com/redskyops/redskyops-www/commit/eef4984c97bc265a3861c87fc83ba18fd4d95e1d)?

jgustie:
FWIW, those references are from annotated tags.

bradbeam:
Definitely cleaner; think I got confused in understanding difference between the two.

.github/workflows/master.yaml
```diff
@@ -15,8 +15,8 @@ jobs:
     steps:
       - name: Check out code
         uses: actions/checkout@v2
-      - name: Unshallow
-        run: git fetch --prune --unshallow
+      - name: Fetch tags
+        run: git fetch --depth=1 origin '+refs/tags/*:refs/tags/*'
```
jgustie, line 19: Can we just leave this like it was so it matches the [documentation it came](https://goreleaser.com/actions/) from?

Makefile
```diff
@@ -15,7 +15,7 @@ GOBIN=$(shell go env GOBIN)
 endif
 
 # Collect version information
-VERSION ?= $(shell git ls-remote --tags origin 'v*' | tail -1 | awk -F/ '{ print $$3 }')-next
+VERSION ?= $(shell git ls-remote --tags --refs origin 'v*' | awk -F/ '{ print $$3 }' | tail -1)-next
```
jgustie, line 18: This 68k Jeremy speaking, why not `tail` first and only `awk` the one line?

.github/workflows/master.yaml
bradbeam: Was torn on switching to tags or adding in `fetch-depth: 0` to save a step.

```
    # Number of commits to fetch. 0 indicates all history.
    # Default: 1
    fetch-depth: '
```

Makefile
bradbeam: I think it was just left over from switching out the change. I can switch back.

.github/workflows/master.yaml
jgustie: You are right, getting the tags is better then `fetch-depth: 0`. I think "new Git" lets you do `git fetch --tags --depth=1` instead of spelling out the refspec.

jgustie (approved):

PR #69, opened 2020-05-08
feat: Add cache step for docker build

bradbeam:

### DIFF
--- a/.github/workflows/master.yaml
+++ b/.github/workflows/master.yaml
@@ -47,6 +47,7 @@ jobs:
           echo "::set-env name=IMG::gcr.io/${{ secrets.GOOGLE_PROJECT_ID }}/redskyops-controller:${TAG}"
           echo "::set-env name=REDSKYCTL_IMG::gcr.io/${{ secrets.GOOGLE_PROJECT_ID }}/redskyctl:${TAG}"
           echo "::set-env name=SETUPTOOLS_IMG::gcr.io/${{ secrets.GOOGLE_PROJECT_ID }}/setuptools:${TAG}"
+          echo "::set-env name=CACHE_IMG::gcr.io/${{ secrets.GOOGLE_PROJECT_ID }}/redskyops-cache:latest"
           echo "::set-env name=PULL_POLICY::Always"
           echo "::set-env name=DOCKER_TAG::${DOCKER_TAG}"
           gcloud --quiet auth configure-docker
--- a/.github/workflows/push.yaml
+++ b/.github/workflows/push.yaml
@@ -21,6 +21,13 @@ jobs:
         uses: actions/setup-go@v2
         with:
           go-version: '1.14'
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -76,13 +77,23 @@ generate: controller-gen
...
+		--target cache \
+		--cache-from=${CACHE_IMG}
+	docker build . -t ${IMG} \
+		--build-arg LDFLAGS='${LDFLAGS}' \
+		--cache-from=${CACHE_IMG}
+	docker build config -t ${SETUPTOOLS_IMG} \
+		--build-arg IMG='${IMG}' \
+		--build-arg PULL_POLICY='${PULL_POLICY}' \
+		--build-arg VERSION='${VERSION}' \
+		--cache-from=${CACHE_IMG}
```
jgustie: You don't need to use the cache image for the setuptools.

.github/workflows/push.yaml
```diff
@@ -33,6 +33,7 @@ jobs:
           echo "::set-env name=IMG::gcr.io/${{ secrets.GOOGLE_PROJECT_ID }}/redskyops-controller:${TAG}"
           echo "::set-env name=REDSKYCTL_IMG::gcr.io/${{ secrets.GOOGLE_PROJECT_ID }}/redskyctl:${TAG}"
           echo "::set-env name=SETUPTOOLS_IMG::gcr.io/${{ secrets.GOOGLE_PROJECT_ID }}/setuptools:${TAG}"
+          echo "::set-env name=CACHE_IMG::gcr.io/${{ secrets.GOOGLE_PROJECT_ID }}/redskyops-cache:latest"
```
jgustie, line 50: This would need to be duplicated in the main workflow.

jgustie:
Can you also pull in the "Cache Go Modules" step from the master workflow? I think I forgot to add that to the pull workflow.

Push the change to remove the `--cache-from` on the setup tools first though: I would like to see the timings strictly from the Docker cache image first.

.github/workflows/master.yaml
```diff
@@ -17,10 +17,6 @@ jobs:
         uses: actions/checkout@v2
       - name: Unshallow
         run: git fetch --prune --unshallow
-      - name: Set up Go
```
bradbeam: https://github.com/actions/virtual-environments/blob/8491d71a0b76ae4a8572a27f683e396269bfa80c/images/linux/ubuntu1804.json#L27-L28 // https://github.com/actions/virtual-environments/blob/master/images/linux/Ubuntu1804-README.md

Looks like go 1.14 is included in the runner by default, so we get to skip a step here.

bradbeam: Eh, added this back in. There's a bit of a lag ( 1.14.1 vs 1.14.2 )

bradbeam:
closing for now since there were no real benefits

PR #71, opened 2020-05-11
Eliminate aggregate rule on cluster role

jgustie:
Currently installation requires elevated privileges (e.g. `cluster-admin`) to create an aggregate rule on the cluster role use assign "patch" permissions to the Red Sky Controller.

This change eliminates the cluster role with the aggregate rule and now generates an explicit cluster role binding along with roles that were previously relying on the aggregation rule.


redskyctl/internal/commander/printer.go
```diff
@@ -462,21 +462,47 @@ func (k *kubePrinter) Header(outputFormat string, column string) string {
...
-	if err := scheme.Convert(obj, u, runtime.InternalGroupVersioner); err == nil {
-		removeCreationTimestamp(u.UnstructuredContent())
-		if k.hideStatus {
-			delete(u.UnstructuredContent(), "status")
+	if err := k.convert(obj, u); err != nil {
+		return k.printer.PrintObj(obj, w)
+	}
+
+	// List conversion is not deep, explicitly convert each item
+	if l, ok := obj.(*corev1.List); u.IsList() && ok {
```
bradbeam: nit, could we return early on this?

bradbeam (approved):

bradbeam (approved):

PR #77, opened 2020-05-13
feat: Introduce build matrix for kube versions

bradbeam:

### DIFF
--- a/.github/workflows/kubematrix.yaml
+++ b/.github/workflows/kubematrix.yaml
@@ -1,79 +1,22 @@
 name: Matrix workflow
 on:
-  push:
+  #push:
+  pull_request:
     branches:
       - master
-    tags:
-      - v*
-
 jobs:
-  build:
-    name: Build and Deploy
-    runs-on: ubuntu-latest
-    env:
-      BUILD_METADATA: build.${{ github.run_number }}
-      GIT_COMMIT: ${{ github.sha }}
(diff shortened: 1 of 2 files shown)


.github/workflows/kubematrix.yaml
```diff
@@ -0,0 +1,122 @@
+name: Main workflow
```
jgustie: Definitely change the name, I think it's confusing the branch restrictions.

jgustie:
I wonder if we can pull the output of a previous build instead of having to rebuild just for this workflow?

bradbeam:
> I wonder if we can pull the output of a previous build instead of having to rebuild just for this workflow?

I think the only way to make this happen is to have the jobs be part of the same workflow. We can add in the matrix build stuff into the master build if you want to go that route. Since it runs on push to master I'm not sure I mind too much on any additional time for this, but I can see the benefit of using the exact same artifacts that would be published.

jgustie:
This looks like it's going to attempt to re-release the product since it's re-deploying everything and it's running on a tag build.

bradbeam:
Should we merge the workflows then?

jgustie:
At a minimum we just shouldn't trigger this workflow on tag pushes (in theory it would have already run on the commit push to master anyway).

jgustie (approved):

PR #81, opened 2020-05-14
Allow the default setup tools image name to be overridden

jgustie:
This is intended for special cases where the controller binaries are deployed without access to the public image repository.

### DIFF
--- a/internal/setup/job.go
+++ b/internal/setup/job.go
@@ -19,6 +19,7 @@ package setup
 import (
 	"encoding/base64"
 	"fmt"
+	"os"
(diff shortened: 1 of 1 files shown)


bradbeam (approved):

bradbeam:
Should we expose this as a flag to redskyctl init as well?

jgustie:
I don't think it's necessary in the tool (you can always run the image yourself without `redskyctl` if you really need to); in this case the only workaround would have been to manually specify the alternate image on every single experiment. We can revisit if it comes up.

PR #82, opened 2020-05-14
chore: Split out tests from build

bradbeam:
Attempt to speed up builds by allowing tests and build to run in parallel

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -7,3 +7,4 @@
 !/go.mod
 !/go.sum
 !/main.go
+!/manager
--- a/.github/workflows/push.yaml
+++ b/.github/workflows/push.yaml
@@ -7,9 +7,6 @@ jobs:
   test:
     name: Unit Tests
     runs-on: ubuntu-latest
-    env:
-      BUILD_METADATA: pr.${{ github.run_number }}
-      GIT_COMMIT: ${{ github.sha }}
     steps:
       - name: Check out code
         uses: actions/checkout@v2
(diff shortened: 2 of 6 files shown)


jgustie:
Can we just remove the `test` dependency from `docker-build` and then call `make test` explicitly instead of duplicating the target?

.github/workflows/push.yaml
```diff
@@ -4,23 +4,57 @@ on:
...
         with:
           go-version: '1.14'
+      - name: Bootstrap
+        run: |
+          TAG=${GITHUB_SHA:0:8}.${GITHUB_RUN_ID}
+          echo "::set-env name=IMG::gcr.io/${{ secrets.GOOGLE_PROJECT_ID }}/redskyops-controller:${TAG}"
+          echo "::set-env name=REDSKYCTL_IMG::gcr.io/${{ secrets.GOOGLE_PROJECT_ID }}/redskyctl:${TAG}"
+          echo "::set-env name=SETUPTOOLS_IMG::gcr.io/${{ secrets.GOOGLE_PROJECT_ID }}/setuptools:${TAG}"
+          echo "::set-env name=PULL_POLICY::Always"
+          gcloud --quiet auth configure-docker
```
jgustie: Is this bootstrap step needed for the tests?

.github/workflows/push.yaml
```diff
@@ -4,23 +4,57 @@ on:
...
       - master
 jobs:
-  build:
-    name: Build and Deploy
+  test:
+    name: Unit Tests
     runs-on: ubuntu-latest
     env:
       BUILD_METADATA: pr.${{ github.run_number }}
       GIT_COMMIT: ${{ github.sha }}
```
jgustie: This env isn't needed for just the tests

.dockerignore
```diff
@@ -7,3 +7,4 @@
 !/go.mod
 !/go.sum
 !/main.go
+!/manager
```
jgustie, line 10: Add this to the `.gitignore` also.

jgustie: ...well add the opposite of this to the Git ignore

jgustie (approved):

PR #84, opened 2020-05-15
Add support for generation of scoped bindings

jgustie:
I have renamed `redskyctl generate bootstrap-cluster-role` to `redskyctl generate controller-rbac` and added two new options: `--ns-selector SELECTOR` and `--include-manager`.

Typically the role of this generator is provide an additional "default" cluster role and cluster role binding so that things "just work" out of the box when you write experiments that only patch Deployments, Stateful Sets and Config Maps.

Now, when you specify `--ns-selector` we will still generate a cluster role, but we will create namespace scoped role bindings instead of a cluster role binding. This allows you to grant the controller permissions to `get` and `patch` the Deployments, Stateful Sets and Config Maps only in certain namespaces (i.e. those matching the specified selector).

Additionally, when `--ns-selector` is being used, the `--include-manager` option can be specified to create an additional role binding in each namespace which binds the `redsky-manager-role` (the role required for doing things like accessing experiments, trials, jobs, and pods).

This new behavior will be integrated into `redskyctl init` in a later PR. Similar functionality will also be added to `redskyctl generate rbac` in a later PR.

### DIFF
--- a/redskyctl/internal/commands/grant_permissions/generator.go
+++ b/redskyctl/internal/commands/grant_permissions/generator.go
@@ -17,6 +17,10 @@ limitations under the License.
 package grant_permissions
 
 import (
+	"bufio"
+	"bytes"
+	"context"
+
 	"github.com/redskyops/redskyops-controller/internal/config"
 	"github.com/redskyops/redskyops-controller/redskyctl/internal/commander"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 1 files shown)


redskyctl/internal/commands/grant_permissions/generator.go
```diff
@@ -67,41 +75,76 @@ func NewGeneratorCommand(o *GeneratorOptions) *cobra.Command {
 func (o *GeneratorOptions) addFlags(cmd *cobra.Command) {
 	cmd.Flags().BoolVar(&o.SkipDefault, "skip-default", o.SkipDefault, "Skip default permissions.")
 	cmd.Flags().BoolVar(&o.CreateTrialNamespaces, "create-trial-namespace", o.CreateTrialNamespaces, "Include trial namespace creation permissions.")
+	cmd.Flags().StringVar(&o.NamespaceSelector, "ns-selector", o.NamespaceSelector, "Bind to matching namespaces.")
```
bradbeam, line 78: I think we'd also want to support selecting namespaces by name.

jgustie: I was thinking about it, but I wasn't sure what it would look like.

bradbeam (approved):

PR #86, opened 2020-05-19
Adjust generate rbac to produce cluster or scoped roles

jgustie:
Kept the core "find patches, generate roles" logic, but changed it so we can generate either:
1. ClusterRole/ClusterRoleBinding (default and current behavior)
2. ClusterRole/RoleBinding (similar to scoped installed for the manager role)
3. Role/RoleBinding (a new "experiment only" combination)

Most of these changes were just around grouping related bits of logic and having the ability to change the combination of object types that gets created.

Also new is that we need to consider the namespace for each experiment read in: if we are creating a Role or RoleBinding we need a namespace (even if it's an empty string which means the namespace will come from `kubectl apply/create` ultimately falling back to `default` if left unspecified). I opted to use the "cluster access" namespace as a default (this is the namespace that would typically get passed to every `kubectl` invocation).

### DIFF
--- a/redskyctl/internal/commands/generate/rbac.go
+++ b/redskyctl/internal/commands/generate/rbac.go
@@ -17,7 +17,9 @@ limitations under the License.
 package generate
 
 import (
+	"fmt"
 	"path/filepath"
+	"regexp"
 	"sort"
 	"strings"
 
@@ -30,11 +32,11 @@ import (
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/rand"
 	"k8s.io/client-go/kubernetes/scheme"
 )
 
-// TODO Determine if this should be exposed as a Kustomize plugin also
-// TODO Instead of it's own plugin, have it be an option on the experiment plugin
+// TODO Instead of it's own generator, have this be an option on the experiment generator
 
 type RBACOptions struct {
 	// Config is the Red Sky Configuration used to generate the role binding
@@ -44,9 +46,11 @@ type RBACOptions struct {
 	// IOStreams are used to access the standard process streams
 	commander.IOStreams
 
-	Filename     string
-	Name         string
-	IncludeNames bool
+	Filename           string
+	Name               string
+	IncludeNames       bool
+	ClusterRole        bool
+	ClusterRoleBinding bool
 
 	mapper meta.RESTMapper
 }
@@ -72,6 +76,8 @@ func NewRBACCommand(o *RBACOptions) *cobra.Command {
 	cmd.Flags().StringVarP(&o.Filename, "filename", "f", o.Filename, "File that contains the experiment to extract roles from.")
 	cmd.Flags().StringVar(&o.Name, "role-name", o.Name, "Name of the cluster role to generate (default is to use a generated name).")
 	cmd.Flags().BoolVar(&o.IncludeNames, "include-names", o.IncludeNames, "Include resource names in the generated role.")
+	cmd.Flags().BoolVar(&o.ClusterRole, "cluster-role", o.ClusterRole, "Generate a cluster role.")
+	cmd.Flags().BoolVar(&o.ClusterRoleBinding, "cluster-role-binding", o.ClusterRoleBinding, "When generating a cluster role, also generate a cluster role binding.")
 
 	_ = cmd.MarkFlagFilename("filename", "yml", "yaml")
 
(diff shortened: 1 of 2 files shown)


redskyctl/internal/commands/generate/rbac.go
```diff
@@ -90,88 +96,174 @@ func (o *RBACOptions) Complete() {
...
+		roleRef.Kind = "ClusterRole"
+	}
+	if roleRef.Name == "" {
+		roleRef.Name = "redsky-patching-"
+		if len(experimentList.Items) == 1 && experimentList.Items[0].Name != "" {
+			roleRef.Name += experimentList.Items[0].Name
 		} else if o.Filename == "-" {
-			// TODO This needs some type of uniqueness
-			clusterRole.Name = "redsky-patching-stdin"
+			roleRef.Name += fmt.Sprintf("stdin-%s", rand.String(5))
```
bradbeam: any particular reason we'd handle this differently than the default ( `roleRef.Name += rand.String(5)` )

redskyctl/internal/commands/generate/rbac.go
```diff
@@ -90,88 +96,174 @@ func (o *RBACOptions) Complete() {
...
+}
+
+func (o *RBACOptions) bindingTargets(experimentList *redskyv1alpha1.ExperimentList) (*rbacv1.RoleRef, *rbacv1.Subject, []string, error) {
+	// Create the role reference
+	roleRef := &rbacv1.RoleRef{APIGroup: "rbac.authorization.k8s.io", Kind: "Role", Name: o.Name}
+	if o.ClusterRole {
+		roleRef.Kind = "ClusterRole"
+	}
+	if roleRef.Name == "" {
+		roleRef.Name = "redsky-patching-"
```
bradbeam: should we define this as a const?

redskyctl/internal/commands/generate/rbac.go
```diff
@@ -90,88 +96,174 @@ func (o *RBACOptions) Complete() {
...
-		Items: []runtime.RawExtension{
-			{Object: clusterRole},
-			{Object: clusterRoleBinding},
-		},
+		// Try to provide an explicit default value (cstr.Namespace can still be "")
+		if ns[""] {
+			if cstr, err := config.CurrentCluster(o.Config.Reader()); err == nil {
+				delete(ns, "")
+				ns[cstr.Namespace] = true
+			}
```
bradbeam: Ultimately a nit, but could we do something like 

```
	var defaultNamespace string
	if cstr, err := config.CurrentCluster(o.Config.Reader()); err == nil {
		defaultNamespace = cstr.Namespace
	}

	ns := make(map[string]struct{}, len(experimentList.Items))
	for _, experiment := range experimentList.Items {
		if experiment.Namespace == "" {
			experiment.Namespace = defaultNamespace
		}
		ns[experiment.Namespace] = struct{}{}
	}
```
Since we don't actually need the value part of the map, I think a `struct{}` would be more appropriate to indicate this is a just a placeholder value with no actual purpose.

redskyctl/internal/commands/generate/rbac.go
```diff
@@ -90,88 +96,174 @@ func (o *RBACOptions) Complete() {
...
+							Namespace: ns,
+						},
+						Subjects: []rbacv1.Subject{*subject},
+						RoleRef:  *roleRef,
+					},
+				})
+		}
+	}
+	return result
+}
```
bradbeam: What about moving the role binding generation out of the switch statement since it's common in both role and cluster role? -- 


```
	case "ClusterRole":
		// Include a single cluster role
		result.Items = append(result.Items, runtime.RawExtension{
			Object: &rbacv1.ClusterRole{
				ObjectMeta: metav1.ObjectMeta{
					Name: roleRef.Name,
				},
				Rules: rules,
			},
		})

		if len(namespaces) > 0 {
			break
		}
		
		// If there are no namespaces, include a single cluster role binding
		result.Items = append(result.Items, runtime.RawExtension{
			Object: &rbacv1.ClusterRoleBinding{
				ObjectMeta: metav1.ObjectMeta{
					Name: roleRef.Name + "binding",
				},
				Subjects: []rbacv1.Subject{*subject},
				RoleRef:  *roleRef,
			},
		})

	case "Role":
		for _, ns := range namespaces {
			// Include a role and role binding for each namespace
			result.Items = append(result.Items,
				runtime.RawExtension{
					Object: &rbacv1.Role{
						ObjectMeta: metav1.ObjectMeta{
							Name:      roleRef.Name,
							Namespace: ns,
						},
						Rules: rules,
					},
				})
		}
	}

	// For each namespace, include a role binding
	for _, ns := range namespaces {
		result.Items = append(result.Items, runtime.RawExtension{
			Object: &rbacv1.RoleBinding{
				ObjectMeta: metav1.ObjectMeta{
					Name:      roleRef.Name + "binding",
					Namespace: ns,
				},
				Subjects: []rbacv1.Subject{*subject},
				RoleRef:  *roleRef,
			},
		})
	}

	return result
```

bradbeam (approved):

redskyctl/internal/commands/generate/rbac.go
jgustie: So much better.

redskyctl/internal/commands/generate/rbac.go
jgustie: This whole block is a bit iffy. I'm going to move it out to a separate function to clean things up a little.

redskyctl/internal/commands/generate/rbac.go
jgustie: Just to help differentiate experiments read in via stdin. I'm actually going to expand the "special files" bit and also look for `/dev/fd/X` (which would be the file name for process substitution).

PR #91, opened 2020-05-20
Add the ability to label an experiment

jgustie:
Three bits of tech debt I'll address next regarding the Experiments API implementation:
1. All of the metadata URLs should use Hungarian notation to avoid conflicts
2. All of the API methods that accept a URL should explicitly return an error if the URL is empty
3. The `api.go` file is too big, it needs to get split up

### DIFF
--- a/redskyapi/experiments/v1alpha1/api.go
+++ b/redskyapi/experiments/v1alpha1/api.go
@@ -239,7 +239,7 @@ type ExperimentMeta struct {
 	Self         string    `json:"-"`
 	Trials       string    `json:"-"`
 	NextTrial    string    `json:"-"`
-	Labels       string    `json:"-"`
+	LabelsURL    string    `json:"-"`
 }
 
 func (m *ExperimentMeta) SetLocation(string) {}
@@ -255,7 +255,7 @@ func (m *ExperimentMeta) SetLink(rel, link string) {
 	case relationNextTrial:
 		m.NextTrial = link
 	case relationLabels:
-		m.Labels = link
+		m.LabelsURL = link
 	}
 }
 
@@ -358,8 +358,6 @@ type ExperimentLabels struct {
 type TrialMeta struct {
 	ReportTrial string `json:"-"`
 	LabelsURL   string `json:"-"`
-
-	// TODO Add "URL" suffixes on links / rename "ReportTrial" to "SelfURL"
 }
 
 func (m *TrialMeta) SetLocation(location string) { m.ReportTrial = location }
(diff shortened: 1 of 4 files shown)


redskyapi/experiments/v1alpha1/api.go
```diff
@@ -776,6 +814,35 @@ func (h *httpAPI) AbandonRunningTrial(ctx context.Context, u string) error {
 	}
 }
 
+func (h *httpAPI) LabelExperiment(ctx context.Context, u string, lbl ExperimentLabels) error {
+	b, err := json.Marshal(lbl)
+	if err != nil {
+		return err
```
andrewhare, line 820: Nitpick for the future: We should consider either doing this:

```go
return fmt.Errorf("failed to marshal JSON: %s", err.Error())
```

or use something like https://github.com/pkg/errors to wrap these kind of errors with more context.

Hopefully with Go 2 this will become part of the stdlib.

andrewhare (approved):

redskyapi/experiments/v1alpha1/api.go
bradbeam: Don't we get something similar to that with [go 1.13](https://blog.golang.org/go1.13-errors)

```
Wrapping errors with %w

As mentioned earlier, it is common to use the fmt.Errorf function to add additional information to an error.

> if err != nil {
>    return fmt.Errorf("decompress %v: %v", name, err)
>}

In Go 1.13, the fmt.Errorf function supports a new %w verb. When this verb is present, the error returned by fmt.Errorf will have an Unwrap method returning the argument of %w, which must be an error. In all other ways, %w is 
identical to %v.

> if err != nil {
>    // Return an error which unwraps to err.
>    return fmt.Errorf("decompress %v: %w", name, err)
> }
Wrapping an error with %w makes it available to errors.Is and errors.As:

> err := fmt.Errorf("access denied: %w", ErrPermission)
> ...
> if errors.Is(err, ErrPermission) 
> ...
```

bradbeam (approved):

redskyapi/experiments/v1alpha1/api.go
andrewhare: We do, but hopefully in Go 2 we will get an API that is closer to what `pkg/errors` gives us.

Either way, adding context to the error helps down the road when you look in the logs and see `EOF` and have no idea where it came from.

jgustie: Yeah, the error handling is basically straight pass through everywhere. I like the simplicity of judiciously using `%w`. I'm not a fan of blindly wrapping everything, but there are definitely some places it would clean things up (e.g. [`IsUnauthorized`](https://github.com/redskyops/redskyops-controller/blob/e472add7287245c1a369aa429416e5ec71347794/redskyapi/experiments/v1alpha1/api.go#L108)).

PR #93, opened 2020-05-21
Red Sky Experiments API cleanup

jgustie:
As previously promised.

This breaks up `api.go` into smaller chunks. Adds the Hungarian notation for URL strings (more uniformly). And it fixes some of the error reporting.

I still am not wrapping any of the errors: the approach I am taking is to return the raw error until I see a message printed out by the CLI that really isn't useful.


jgustie:
Note that for the most part `api.go` just had things removed (the `Error` changed a little); the `experiment.go` and `trial.go` are just wholesale what moved out of `api.go` (the Hungarian notation was added). The `http.go` is where the real changes were (to make the methods more consistent and to better handle errors).

My favorite part about this change is that we finally capture the error messages from the server.

jgustie:
👋

redskyapi/experiments/v1alpha1/trial.go
```diff
@@ -0,0 +1,146 @@
...
+		strs := make([]string, len(p.Status))
+		for i := range p.Status {
+			strs[i] = string(p.Status[i])
+		}
+		q.Add("status", strings.Join(strs, ","))
+	}
+	if len(p.LabelSelector) > 0 {
+		ls := make([]string, 0, len(p.LabelSelector))
+		for k, v := range p.LabelSelector {
+			ls = append(ls, fmt.Sprintf("%s=%s", k, v))
```
bradbeam, line 127: not sure how much of this is carry over, but should we use https://golang.org/pkg/net/url/#QueryEscape instead of `fmt.Sprintf`?

bradbeam (approved):

redskyapi/experiments/v1alpha1/trial.go
jgustie: That would double encode, i.e. this function would return `labelSelector=foo%253Dbar` instead of just `labelSelector=foo%3Dbar`.

This output gets dropped into `URL.RawQuery` which does the right thing (which isn't really the right thing because technically `=` doesn't need to be encoded here but it's better to error on the side of caution because some URL parsers are just garbage).

PR #95, opened 2020-06-03
v1beta1

jgustie:
This massive PR adds v1beta1 as a second version for our CRD resources. It provides backwards compatibility with the old resource versions. It does not leverage webhook based conversions (though the necessary interfaces _are_ implemented should we go that route in the future).

I will comment on the interesting parts of the code to make it easier to follow exactly what changed. At a high level I used the IDE to refactor the `v1alpha1` to `v1beta1` and then copied the structs back to `v1alpha1`. @bradbeam provided the `conversion-gen` (replacing my initial hand-written implementation): this had the benefit of allowing us to use Kube API Machinery based conversions else where in the code.

To provide compatibility with old resources, the `client-go` code was extended with a `NegotiatedSerializer` that inspects fetched representations of `redskyops.dev` resources to identify representations which require conversion. If a resource does require conversion it is done in the REST client, before the controller-runtime code ever gets to see it. This provides a lazy migration strategy through the controller (i.e. only updated resources ever get migrated).

Finally, changes were made to the new `v1beta1` version: namely the `template` fields were renamed to provide more context and the several fields were moved to the trial status (in particular those which are not expected to be configured by the user but are used internally by the controller).

The approach to convert objects at decode time is a hack. It was done in large part to delay the need for an API server webhook.


Makefile
```diff
@@ -106,8 +108,26 @@ else
...
+CONVERSION_GEN=$(shell which conversion-gen)
+endif
+
 # Generate CLI and API documentation
 generate-docs:
 	rm -rf docs/redskyctl docs/api
 	go run -ldflags '$(LDFLAGS)' redskyctl/main.go docs --directory docs/redskyctl
-	go run -ldflags '$(LDFLAGS)' redskyctl/main.go docs --directory docs/api --doc-type api
+	go run -ldflags '$(LDFLAGS)' redskyctl/main.go docs --directory docs/api/v1beta1 --source api/v1beta1 --doc-type api
+	go run -ldflags '$(LDFLAGS)' redskyctl/main.go docs --directory docs/api/v1alpha1 --source api/v1alpha1 --doc-type api
```
jgustie, line 133: This change will require a subsequent update to the documentation build.

api/v1alpha1/experiment_conversion.go
```diff
@@ -0,0 +1,61 @@
...
+
+package v1alpha1
+
+import (
+	"github.com/redskyops/redskyops-controller/api/v1beta1"
+	"k8s.io/apimachinery/pkg/conversion"
+	conv "sigs.k8s.io/controller-runtime/pkg/conversion"
+)
+
+var _ conv.Convertible = &Experiment{}
```
jgustie, line 25: This makes our types convertible via webhook (once we choose to add one). Using a transient scheme ensures we do not call generated conversion methods inappropriately. Also, the way the generated conversion methods work, as soon as there is a conflict between the objects that cannot be generated, it drops the method and forces you to implement it yourself.

internal/controller/negotiation.go
```diff
@@ -0,0 +1,234 @@
...
+	return &ConversionDecoder{Decoder: c.NegotiatedSerializer.DecoderToVersion(serializer, gv), scheme: c.scheme}
+}
+
+// ConversionDecoder is a decoder that decodes with the representation migration hack.
+type ConversionDecoder struct {
+	runtime.Decoder
+	scheme *runtime.Scheme
+}
+
+func (c *ConversionDecoder) Decode(data []byte, defaults *schema.GroupVersionKind, into runtime.Object) (runtime.Object, *schema.GroupVersionKind, error) {
```
jgustie, line 70: This method is invoked on each raw response coming back from the Kube API server. Typically, because of the way the controller-runtime cache works, this will we a `List` object (i.e. `ExperimentList`). For our objects this will check expected non-zero values before deserializing a second time (using the old version), thus allowing us to confirm that an object requires conversion to a newer representation. By injecting this logic here we ensure the controller never actually sees the old representations.

internal/controller/negotiation.go
```diff
@@ -0,0 +1,234 @@
+/*
```
bradbeam, line 1: Can be in a follow up PR, but could we get a test added for this ?

bradbeam (approved):

PR #102, opened 2020-06-05
feat: Deprecate setup tools on init

bradbeam:
This introduces the ability to generate the installation manifests internally via redskyctl/kustomize
so we no longer need to rely on execing into the setup tools pod.


redskyctl/internal/commands/initialize/generator.go
```diff
@@ -49,49 +52,22 @@ func NewGeneratorCommand(o *GeneratorOptions) *cobra.Command {
...
+	cmd.Flags().StringVar(&o.Image, "image", kustomize.BuildImage, "Specify the controller image to use.")
+
 	commander.ExitOnError(cmd)
 	return cmd
 }
 
 func (o *GeneratorOptions) generate(ctx context.Context) error {
-	// Read the initial information from the configuration
-	r := o.Config.Reader()
-	ctrl, err := config.CurrentController(r)
```
jgustie, line 59: The namespace for initialization needs to come out of the configuration, i.e. `ctrl.Namespace` is where `redsky-system` comes from.

redskyctl/internal/commands/initialize/generator.go
```diff
@@ -49,49 +52,22 @@ func NewGeneratorCommand(o *GeneratorOptions) *cobra.Command {
...
-	args = append(args, "--restart", "Never", "--attach")
-
-	// Quietly remove the pod when we are done
-	args = append(args, "--rm", "--quiet")
-
-	// Use the image embedded in the code
-	args = append(args, "--image", setup.Image)
-	// TODO We may need to overwrite this for offline clusters
-	args = append(args, "--image-pull-policy", setup.ImagePullPolicy)
+	fmt.Println(string(yamls))
```
jgustie: This should `Fprintln` to `o.Out`.

config/default/manager_patch.yaml
```diff
@@ -0,0 +1,14 @@
...
+  namespace: system
+spec:
+  template:
+    spec:
+      containers:
+      - name: manager
+        envFrom:
+        - secretRef:
+            name: redsky-manager
+            optional: true
```
jgustie: I would rather not have this moved into `config/default` as it that would include it into other downstream kustomizations.

bradbeam: Should we try to inject this conditionally if there is a complete redskyctl config file?

jgustie: Sure: this is here so we could have one static manifest for both the OSS and Enterprise editions, but having it optional has caused some unique timing related issues that are hard to debug.

Makefile
```diff
@@ -38,8 +39,12 @@ manager: generate fmt vet
...
 # Build tool binary using GoReleaser in a local dev environment (in CI we just invoke GoReleaser directly)
 tool:
-	BUILD_METADATA=${BUILD_METADATA} SETUPTOOLS_IMG=${SETUPTOOLS_IMG} PULL_POLICY=${PULL_POLICY} REDSKYCTL_IMG=${REDSKYCTL_IMG} \
-	goreleaser release --snapshot --skip-sign --rm-dist
+	BUILD_METADATA=${BUILD_METADATA} \
+	SETUPTOOLS_IMG=${SETUPTOOLS_IMG} \
+	PULL_POLICY=${PULL_POLICY} \
+	REDSKYCTL_IMG=${REDSKYCTL_IMG} \
+	IMG=${IMG} \
+	goreleaser release --snapshot --skip-sign --rm-dist --debug
```
jgustie: Is adding `--debug` here intentional?

Makefile
```diff
@@ -81,7 +87,7 @@ generate: controller-gen conversion-gen
 docker-build: test docker-build-ci
 
 # Build the docker images
-docker-build-ci:
+docker-build-ci: manifests
```
jgustie: Should we be committing the generated files like we do with the rest of the generated code so we do not need to regenerate manifests on every build?

redskyctl/internal/commands/initialize/generator.go
```diff
@@ -49,49 +52,40 @@ func NewGeneratorCommand(o *GeneratorOptions) *cobra.Command {
...
-	args = append(args, "--overrides", `{"spec":{"automountServiceAccountToken":false}}`)
+	auth, err := config.CurrentAuthorization(r)
+	if err != nil {
+		return err
+	}
 
-	// Overwrite the "redsky-system" namespace
-	args = append(args, "--env", "NAMESPACE="+ctrl.Namespace)
+	apiEnabled := false
+	if auth.Credential.TokenCredential != nil {
```
jgustie, line 74: You should check `auth.Credential.ClientCredential` as well.

jgustie (approved):

cmd/generator/main.go
```diff
@@ -0,0 +1,85 @@
...
+
+	flag "github.com/spf13/pflag"
+	"sigs.k8s.io/kustomize/api/filesys"
+	"sigs.k8s.io/kustomize/api/krusty"
+)
+
+func main() {
+	var (
+		path          = flag.String("path", ".", "path of kustomization.yaml")
+		outputPackage = flag.String("package", "generated", "output package name")
```
jgustie, line 42: This isn't configurable because of the reference to the `Asset` struct; you would need to conditionally add an `import` statement to keep it configurable.

Makefile
jgustie: I know we don't use this in CI, but in testing out the code signing, it definitely prints sensitive information.

bradbeam: It shouldn't be there :)

Makefile
bradbeam: initially it was because I wanted the build version information from the CI run to get injected when we built the manifests. but after switching over to using kustomize, im not sure it's needed anymore. I'll double check and drop it if we don't need to

cmd/generator/main.go
bradbeam: There's a dependency on having an Asset type defined, but the package name itself doesn't matter ( we're generating it for `package kustomize` ). We could probably move the `redskyctl/internal/kustomize/assets.go` bit into this command to generate the assets struct as well, but I'm not sure it's worth tossing all of that go code inside of a string literal or the like.

PR #106, opened 2020-06-11
Build improvements

jgustie:
This is really a combination of build improvements for release builds:
1. GoReleaser Action update
2. Release images now go to Docker Hub (CI images continue to go to GCR)
3. Code signing and notarization for macOS

For the code signing, I originally wanted to use gon, however between the inconsistent examples and the pain of having to mass duplicate a bulk of the GoReleaser configuration, I opted instead to add two scripts. Both scripts are basically no-ops if we are not running on macOS or if the binary (or archive) in question is not targeted for macOS. The code signing script runs against every single "master" build (we do not sign PRs), however just because the binaries are signed, it does not mean GateKeeper will let them through. The benefit of this approach is that the rest of our build pipeline remains in-tact (e.g. we do not need to switch to `.zip` archives), once the binary itself is signed it doesn't matter how it gets packaged. Notarization is the process of submitting the signed binary to Apple: they do not accept tarballs so we re-package to a Zip file before sending it off. Notarization is configured as a "signing" step in GoReleaser which means it is essentially skipped everywhere but on release builds.

### DIFF
--- a/hack/codesign.sh
+++ b/hack/codesign.sh
@@ -1,0 +1,45 @@
+#!/usr/bin/env bash
+set -eu
+
+# Code Signing
+# ============
+# This script facilitates code signing of binaries for macOS. This script conditionally signs a binary
+# under the following conditions:
+# 1. The binary path contains "darwin" (this prevents duplicating GoReleaser build configurations)
+# 2. The `security` and `codesign` programs are available (this allows the script to be run on Linux)
+# 3. The P12 key pair used as the signing identity is available (allows the script to be run on non-release builds)
+
+# Verify, but exit normally to ensure the conditional nature of the checks
+FILE="${1:?missing file argument}"
+[ "$(basename "$(dirname "$FILE")")" == "redskyctl_darwin_amd64" ] || { echo >&2 "skipping code signing on file '$1'"; exit; }
+command -v security >/dev/null 2>&1 || { echo >&2 "skipping code signing, security not present"; exit; }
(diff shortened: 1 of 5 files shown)


hack/codesign.sh
```diff
@@ -0,0 +1,45 @@
...
+# This script facilitates code signing of binaries for macOS. This script conditionally signs a binary
+# under the following conditions:
+# 1. The binary path contains "darwin" (this prevents duplicating GoReleaser build configurations)
+# 2. The `security` and `codesign` programs are available (this allows the script to be run on Linux)
+# 3. The P12 key pair used as the signing identity is available (allows the script to be run on non-release builds)
+
+# Verify, but exit normally to ensure the conditional nature of the checks
+FILE="${1:?missing file argument}"
+[ "$(basename "$(dirname "$FILE")")" == "redskyctl_darwin_amd64" ] || { echo >&2 "skipping code signing on file '$1'"; exit; }
+command -v security >/dev/null 2>&1 || { echo >&2 "skipping code signing, security not present"; exit; }
```
bradbeam, line 15: Should we flex based on `uname -s` instead?

```
if [ $(uname -s) != "Darwin" ]; then exit; fi
```

bradbeam (approved):

hack/codesign.sh
jgustie: I figured it was better to test that the tools were there, e.g. a Mac without Xcode installed. The `security` and `ditto` commands should always be present, but I think `codesign` and definitely `xcrun` come in from Xcode. Also, I check for `jq` to parse the `xcrun` output (which is installed on the GitHub Action runners).

PR #113, opened 2020-06-23
Secret generation improvements

jgustie:
Several improvements to secret generation highlighted by the current authorization mess. Again, the main goal is to unblock developers; however we found that the errors were not helpful (they still are pretty vague but we do not get anything better from the server right now).

### DIFF
--- a/internal/oauth2/registration/registration.go
+++ b/internal/oauth2/registration/registration.go
@@ -175,10 +175,9 @@ func doRegistrationRoundTrip(ctx context.Context, req *http.Request, src oauth2.
 
 	if code := r.StatusCode; code < 200 || code > 299 {
 		responseError := &ClientRegistrationErrorResponse{}
-		if err := json.Unmarshal(body, responseError); err != nil {
-			if t, _, err := mime.ParseMediaType(r.Header.Get("Content-Type")); err == nil && t != "application/json" {
-				return nil, fmt.Errorf("registration: invalid response body type; expected JSON, got: %s", t)
-			}
+		if t, _, err := mime.ParseMediaType(r.Header.Get("Content-Type")); err != nil || t != "application/json" {
+			return nil, fmt.Errorf("registration: unexpected server error (%v - %s)", code, http.StatusText(code))
+		} else if err := json.Unmarshal(body, responseError); err != nil {
 			return nil, fmt.Errorf("registration: invalid response: %v", err)
 		}
 		if responseError.ErrorCode == "" {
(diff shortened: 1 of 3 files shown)


redskyctl/internal/commands/authorize_cluster/generator.go
```diff
@@ -76,15 +76,31 @@ func NewGeneratorCommand(o *GeneratorOptions) *cobra.Command {
 		RunE: commander.WithContextE(o.generate),
 	}
 
+	// Provide a more meaningful default client name if possible
+	if o.ClientName == "" {
```
bradbeam, line 81: is it worth moving this into `generate()` so we can `s/context.TODO()/cmd.Context()/`?

internal/oauth2/registration/registration.go
```diff
@@ -175,10 +175,9 @@ func doRegistrationRoundTrip(ctx context.Context, req *http.Request, src oauth2.
 
 	if code := r.StatusCode; code < 200 || code > 299 {
 		responseError := &ClientRegistrationErrorResponse{}
-		if err := json.Unmarshal(body, responseError); err != nil {
-			if t, _, err := mime.ParseMediaType(r.Header.Get("Content-Type")); err == nil && t != "application/json" {
-				return nil, fmt.Errorf("registration: invalid response body type; expected JSON, got: %s", t)
-			}
+		if t, _, err := mime.ParseMediaType(r.Header.Get("Content-Type")); err != nil || t != "application/json" {
+			return nil, fmt.Errorf("registration: unexpected server error (%v - %s)", code, http.StatusText(code))
+		} else if err := json.Unmarshal(body, responseError); err != nil {
```
bradbeam: this seems like it could be broken up to help readability -- 

```
if t, _, err := mime.ParseMediaType(r.Header.Get("Content-Type")); err != nil || t != "application/json" {
 			return nil, fmt.Errorf("registration: unexpected server error (%v - %s)", code, http.StatusText(code))
}

if err := json.Unmarshal(body, responseError); err != nil {
...
```

bradbeam (approved):

redskyctl/internal/commands/authorize_cluster/generator.go
jgustie: I put it here so we could include the answer as the default value for the CLI flag (i.e. it shows up in `--help`). The main thing attached to the `cmd.Context()` is the UA strings for HTTP requests which we don't need here. I used `TODO()` instead of `Background()` because the usage better fits the description.

That being said, when I moved it, it still uses `cfg.Kubectl` which is going to be problematic because we haven't hit `preRun` yet (nor will we for `--help`) so the `cfg` hasn't been loaded yet. I'll change it to exec `kubectl` directly in this case.

bradbeam (approved):

PR #114, opened 2020-06-23
Add a wait option to init

jgustie:


bradbeam:
I'm torn on adding a flag for this or just moving the kubectl wait command to hack/integration.sh

jgustie:
I actually did that first, I figured if it was necessary for our automation it would probably be necessary in other contexts. Having it as a flag here also gives us the ability to add additional waits in the future without breaking the contract.

bradbeam (approved):

PR #117, opened 2020-06-26
fix: conditionally load image to kind

bradbeam:
This should allow us to fetch images from a registry during the
matrix builds.


jgustie (approved):
Let's get this in first so the builds pass

PR #119, opened 2020-06-26
refactor: Allow kustomize to be more composable

bradbeam:
This removes the hardcoded restriction on using embedded assets only and
exposes WithAssets and WithPatches to allow our kustomize wrapper to be
more reusable.


bradbeam:
Updated this to not be dependent on other changes, should be good to review.

jgustie (approved):

PR #120, opened 2020-06-30
feat: Add support for exporting trial by label

bradbeam:
This also changes the UX slightly in that youll need to specify -l or -n
for the label or name selector.

will rebase off #110 when that's ready

```
redskyctl export -f $recipes/postgres/experiment.yaml -l brad=true
...
spec:
...
        resources:
          limits:
            cpu: 810m
            memory: 1056Mi
          requests:
            cpu: 810m
            memory: 1056Mi
...
```


redskyctl/internal/commands/export/export.go
```diff
@@ -46,8 +46,10 @@ func NewCommand(cfg config.Config) *cobra.Command {
 		RunE:   ec.Run,
 	}
 
-	cmd.Flags().StringVarP(&ec.filename, "filename", "f", "", "path to experiment file")
-	cmd.MarkFlagRequired("filename")
+	cmd.Flags().StringVarP(&ec.filename, "experiment", "f", "", "path to experiment file")
+	cmd.Flags().StringVarP(&ec.trialName, "trialname", "t", "", "name of trial (ex. postgres-123)")
+	cmd.Flags().StringToStringVarP(&ec.label, "label", "l", nil, "label selector for a trial in k=v format (ex. best=true)")
+	cmd.MarkFlagRequired("experiment")
```
jgustie: I'm jumping ahead of #110.

The flag names should be `--filename/-f`, `--selector/-l` for consistency. Also, object names are not typically given option names, but rather accepted as arguments (all of the "parse names" nonsense to ensure that level of consistency with expectations for users of `kubectl`).

PR #125, opened 2020-07-20
Client registration does not send authorization

jgustie:
This also fixes error reporting to better understand why fetching the client metadata fails.


jgustie:
FWIW this is what makes it possible to reuse `client_id` and `client_secret` without regenerating them every time. I always claimed it was working but until I had a server that supported it, I guess I could never really verify that.

bradbeam (approved):

PR #127, opened 2020-07-23
feat: Introduce metrics.URL parameter

bradbeam:
This introduces the metrics.URL parameter to allow a more explicit
targeting of a remote data source.

Decided to implement a local http test server for fetching the prometheus metric so we can remove the external dependency.


andrewhare (approved):

api/v1beta1/experiment_types.go
```diff
@@ -117,6 +117,10 @@ type Metric struct {
 	Port intstr.IntOrString `json:"port,omitempty"`
 	// URL path component used to collect the metric value from an endpoint (used as a prefix for the Prometheus API)
 	Path string `json:"path,omitempty"`
+	// URL to query for fetching metrics.
```
jgustie, line 120: Do we want to just remove these from `v1beta1`?

jgustie: (that didn't highlight properly, by "these" I meant all the other URL like fields)

bradbeam: Yeah we can, wanted to try to keep the scope of this minimal to make it easier to review :)

jgustie (approved):

jgustie:
I'm waiting to get client registration deployed, at which time 1.6.1 will get released, can we hold off on merging until then? If it's urgent I can release the patch version off a branch.

PR #130, opened 2020-07-30
fix: Update manifests to include url parameter

bradbeam:
Missed regenerating the manifests in #127 


bradbeam:
need to find a good way to catch this

jgustie (approved):
We need to stop committing the `zz_generated.output.go` otherwise the repository size is going to skyrocket.

bradbeam:
> We need to stop committing the zz_generated.output.go otherwise the repository size is going to skyrocket.

as in have this file get generated during build time?

jgustie:
Correct, otherwise we get saddled with a 100K change set in Git for every 20 byte change to a configuration manifest. Those large change sets are unavoidable in history and the Git repository size balloons.

andrewhare:
Can you add it to the gitignore for now?

bradbeam:
I think we tried it at one point and backed off because of the additional time it took.. or was that `conversion_gen`?

I'll update this pr and give it a shot.

bradbeam:
Updated and I think we're still in business.

jgustie (approved):

PR #132, opened 2020-08-14
refactor: Migrate from ObjectReference to PartialObjectMetadata

bradbeam:
Upstream documentation discourages the use of corev1.ObjectReference.
Our usage for ObjectReference can be better represented by PartialObjectMetadata.

ref: https://godoc.org/k8s.io/api/core/v1#ObjectReference


jgustie:
I vote for closing this, though I do think changing the `ExperimentRef` field to a new type (either a plain string in the form `{namespace}/{name}` or a `NamespacedName`) makes sense.

PR #133, opened 2020-08-14
Digest the tenant ID before display from check config

jgustie:


redskyctl/internal/commands/check/config.go
```diff
@@ -78,7 +79,8 @@ func (o *ConfigOptions) checkConfig(ctx context.Context) error {
 		return err
 	}
 	if tenant != "" {
-		_, _ = fmt.Fprintf(o.Out, "Success, configuration is valid for tenant '%s'.\n", tenant)
+		th := sha1.Sum([]byte(tenant))
```
bradbeam, line 82: nit, does it make sense to have this done in `tenantID` ?

bradbeam (approved):

PR #134, opened 2020-08-14
Change "not activated" to "not ready"

jgustie:


jgustie:
Still a draft, verbiage may change. Also, this should be coupled with a change to [the "not activated" page](https://github.com/redskyops/redskyops-www/blob/staging/content/api/auth_not_activated.md).

redskyapi/experiments/v1alpha1/http.go
```diff
@@ -426,7 +426,7 @@ func newError(t ErrorType, resp *http.Response, body []byte) error {
 		case http.StatusPaymentRequired:
 			err.Type = ErrUnauthorized
 			if err.Message == "" {
-				err.Message = "account is not activated"
+				err.Message = "your account is not ready, please check https://app.carbonrelay.io/ for more details"
```
bradbeam: do we need to flex on the url? ( diff between prod/dev ? )

redskyctl/internal/commands/login/login.go
```diff
@@ -288,9 +288,9 @@ func (o *Options) generateCallbackResponse(w http.ResponseWriter, r *http.Reques
 		http.Error(w, http.StatusText(status), status)
 	default:
 		// TODO Better detection of this error
-		if status == http.StatusInternalServerError && err != nil && err.Error() == "account is not activated" {
+		if status == http.StatusInternalServerError && err != nil && strings.HasPrefix(err.Error(), "your account is not ready") {
```
bradbeam: Should we make an actual error type for this, `api.errAccountNotReady` ?

redskyapi/experiments/v1alpha1/http.go
jgustie: It might be nice to change the `login` command to use `https://staging.redskyops.dev/api/auth_success` and `https://app.carbonrelay.dev/` for non-release builds, but in general the CLI is abstracted from the backend deployment environments.

jgustie:
Right now the only way to sign up without triggering a tenant creation is to perform a social login via the CLI. These changes would make it difficult to accommodate tenants with multiple users. Will revisit once the user workflows are sorted out on the backend.

PR #135, opened 2020-08-17
chore: Add additional tests to kustomize

bradbeam:
Added a test to ensure withAPI patches the deployment with envFrom secret.


jgustie:
We should not be checking for client credentials, unless there is a very special case `redskyctl` will only be using an authorization code grant to access client registration.

bradbeam:
Going to reopen with just the first commit

jgustie (approved):

PR #137, opened 2020-08-18
refactor: Create patch package

bradbeam:
This moves the functionality of RenderTemplate and CreatePatchOperation to
a separate pacakge to facilitate reuse.

### DIFF
--- a/internal/patch/patch_test.go
+++ b/internal/patch/patch_test.go
@@ -1,0 +1,215 @@
+/*
+Copyright 2020 GramLabs, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package patch
+
+import (
+	"fmt"
+	"strings"
+	"testing"
+
+	redsky "github.com/redskyops/redskyops-controller/api/v1beta1"
+	"github.com/redskyops/redskyops-controller/internal/template"
+	"github.com/stretchr/testify/assert"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestPatch(t *testing.T) {
+	te := template.New()
+
+	trial := &redsky.Trial{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "mytrial",
+			Namespace: "default",
+		},
+	}
+
+	patchMeta := `apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: myapp
+  namespace: default`
+
+	patchSpec := `spec:
+        template:
+          spec:
+            containers:
+            - name: postgres
+              imagePullPolicy: Always`
+
+	fullPatch := patchMeta + "\n" + patchSpec
+
+	jsonPatch := `[
+    {
+     "op": "replace",
+     "path": "/spec/template/spec/containers/0/imagePullPolicy",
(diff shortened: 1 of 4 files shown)


internal/patch/legacy.go
```diff
@@ -0,0 +1,88 @@
...
+func CreatePatchOperation(t *redsky.Trial, p *redsky.PatchTemplate, ref *corev1.ObjectReference, data []byte) (*redsky.PatchOperation, error) {
+	// If the patch is effectively null, we do not need to evaluate it
+	if len(data) == 0 || string(data) == "null" {
+		return nil, nil
+	}
+
+	// If the patch is for the trial job itself, it cannot be applied
+	// (since the job won't exist until well after patches are applied)
+	if trial.IsTrialJobReference(t, ref) && p.Type != redsky.PatchStrategic {
+		return nil, fmt.Errorf("trial job patch must be a strategic merge patch")
```
jgustie: You need to set attempts remaining to 0 if it is a trial job so we do not actually attempt to apply the patch before the job exists.

internal/patch/legacy.go
```diff
@@ -0,0 +1,88 @@
...
+func RenderTemplate(te *template.Engine, t *redsky.Trial, p *redsky.PatchTemplate) (*corev1.ObjectReference, []byte, error) {
+	// Render the actual patch data
+	data, err := te.RenderPatch(p, t)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// Determine the reference, possibly extracting it from the rendered data
+	ref := &corev1.ObjectReference{}
+	switch {
```
jgustie: I think this was more readable as an `if-else` i.e. "if there is a target ref, deep copy it, else if is an SMP extract the reference from the patch".

internal/patch/legacy.go
```diff
@@ -0,0 +1,88 @@
...
+	case p.TargetRef != nil:
+		p.TargetRef.DeepCopyInto(ref)
+	case p.Type == redsky.PatchStrategic, p.Type == "":
+		m := &struct {
+			metav1.TypeMeta   `json:",inline"`
+			metav1.ObjectMeta `json:"metadata,omitempty"`
+		}{}
```
jgustie: `m := &metav1.PartialObjectMetadata{}` ?!?

internal/patch/patch_test.go
```diff
@@ -0,0 +1,169 @@
...
+	TargetRef corev1.ObjectReference `json:"targetRef"`
+	// The patch content type, must be a type supported by the Kubernetes API server
+	PatchType types.PatchType `json:"patchType"`
+	// The raw data representing the patch to be applied
+	Data []byte `json:"data"`
+	// The number of remaining attempts to apply the patch, will be automatically set
+	// to zero if the patch is successfully applied
+	AttemptsRemaining int `json:"attemptsRemaining,omitempty"`
+}
+*/
```
jgustie: ?

bradbeam: Just sloppy; had it here for reference.

internal/patch/legacy.go
bradbeam: didn't want to cross streams for this.

internal/patch/legacy.go
bradbeam: Why not just ignore any patches that reference the trial job?

internal/patch/legacy.go
bradbeam: I think this might be an agree to disagree bit. I'd really like to find a way to simplify this chunk of logic, but that's out of scope for this round.

internal/patch/legacy.go
jgustie: Attempts remaining is the number of times that we will try to send the patch to Kube API: the trial job is not patchable via the API (everything we would want to change would either be read only or would spin up a new pod). For the trial job, we need to apply the patch before we even create the object.

Setting the attempts remaining to zero means we never try to submit the patch to the Kube API, while keeping a copy of the actual evaluated patch that we can [apply programmatically](https://github.com/redskyops/redskyops-controller/blob/d0677cc446523a3aaa5522a9d38164ed829642de/internal/trial/job.go#L107) to the trial job before it gets created.

jgustie (approved):

PR #143, opened 2020-08-25
docs: Revamp the quickstart guide

bradbeam:
deployment preview here: https://deploy-preview-39--redskyops.netlify.app/docs/quickstart/


docs/quickstart.md
```diff
@@ -1,70 +1,163 @@
-# Quick Start
+# Postgres
```
jgustie: I feel like the top header should always match the file name. Maybe that's just me.

docs/quickstart.md
```diff
@@ -1,70 +1,163 @@
...
 
-## Prerequisites
+If you encounter any issues or require assistance, please reach out to us via [slack][rso slack] or [contact us][contact us form].
 
-You must have a Kubernetes cluster. Additionally, you will need a local configured copy of `kubectl`. The Red Sky Ops Tool will use the same configuration as `kubectl` (usually `$HOME/.kube/config`) to connect to your cluster.
+## Prerequisites
 
-A local install of [Kustomize](https://github.com/kubernetes-sigs/kustomize/releases) (v3.1.0+) is required to build the objects for your cluster.
+- Kubernetes cluster
+  - A typical [minikube][minikube] cluster will be sufficient. The experiment may consume up to 4vcpu and 4GB memory
```
jgustie, line 15: `[minikube][]` ?

docs/quickstart.md
```diff
@@ -1,70 +1,163 @@
...
-This is brief guide to get you up and running with Red Sky Ops as quickly as possible.
+The purpose of this guide is to get you started with Red Sky Ops.
+In this example, you will use Red Sky Ops to tune the memory and cpu requirements for Postgres.
+We will cover:
+- Initializing Red Sky Ops in your cluster
+- Deploying the Postgres experiment to your cluster
+- Running trials using machine learning (ML) to determine the best configuration
 
-## Prerequisites
+If you encounter any issues or require assistance, please reach out to us via [slack][rso slack] or [contact us][contact us form].
```
jgustie: "Slack" instead of "slack".

Instead of the corporate "contact us" form, should we just have a `mailto:support@carbonrelay.com` ?

docs/quickstart.md
```diff
@@ -1,70 +1,163 @@
...
+```sh
+secret/postgres-secret deleted
+service/postgres deleted
+deployment.apps/postgres deleted
+experiment.redskyops.dev/postgres-example deleted
+```
+
 
-## Next Steps
+[install redskyctl]: https://redskyops.dev/docs/install/
```
jgustie: This breaks following the links in the GitHub repository. The existing pattern of relative links works both in both the GitHub source view and on the published website.

docs/quickstart.md
```diff
@@ -1,70 +1,163 @@
...
+- `ClusterRole` and `ClusterRoleBinding` for the controller service account
+- `redsky-controller-manager` deployment
+- an authorization secret for the controller
 
-Once you have the Red Sky Ops Tool you can initialize the manager in your cluster:
+Initialize your cluster with the following command:
 
 ```sh
-$ redskyctl init
+$ redskyctl init --wait
```
jgustie: I don't think `--wait` is necessary for interactive usage; it's should only be necessary for scripts.

jgustie: Same for checking if the manager is running, these feel like trouble shooting steps. In practice all this happens quick enough that a user following a guide would never encounter the state where something wasn't ready (i.e. we should write to the success case, and if the success rate is too low then we need to fix it).

docs/quickstart.md
```diff
@@ -1,70 +1,163 @@
...
+
+This is where you can review the trials that have taken place and decide which parameter makes the most appropriate trade-off.
+There is a filter on the upper right hand side where you can narrow your results to show only the `best` trials.
+These are the trials with the best possible combination of cost and duration, meaning if you hold cost fixed at a given value, we have found the best possible duration.
+You can choose any `best` configuration that works for your cost and duration preferences.
+Often you find that since the metrics are not linearly related, there are some points that give a bigger boost in duration at a similar cost, so it is useful to explore all of the options to find the best one for you.
 
-After the trial is complete, you will be able to view the parameters and the metrics generated from the trial. The metrics can be used to gauge how effective the used parameters were.
+Congratulations! You just ran your first experiment.
+Now that you've got an idea of the basics, you can learn more about the experiment lifecycle [here][lifecycle] or learn more about experiment concepts [here][concepts].
```
jgustie: I'm going on the record as not being a huge fan of "here" link text, e.g. instead of `lifecycle [here][lifecycle]` just use `[lifecycle](lifecycle.md)` (or `[lifecycle][]` if the reference is fixed).

docs/quickstart.md
```diff
@@ -1,70 +1,163 @@
...
 
-If you are planning to create the postgres experiment from this guide, a [minikube](https://kubernetes.io/docs/setup/learning-environment/minikube/) cluster will be sufficient.
+## Initialize the Red Sky Ops Manager
 
-## Install the Red Sky Ops Tool
+After installing [redskyctl][install redskyctl] and creating an [account][rso account], log in to your Red Sky Ops account:
+```sh
+$ redskyctl login
+```
+You should see output similar to this:
```
jgustie: I get the need for positive feedback, but this is the same reason professional documentation writers do not like putting images into their documentation: it's too easy for the output to change thus ruining any established trust you previously gained by illustrating the output (i.e. "That's not exactly what I see, did I do it wrong?").

Maybe as a compromise we could consolidate the two blocks (without the "you should see") and fix preformatted blocks to use the `console` highlighter (instead of `sh`). The `console` highlighter is meant for mimicking shell sessions so it deals with the "prompt" the "command" and the "output".

docs/quickstart.md
```diff
@@ -1,70 +1,163 @@
...
+```sh
+namespace/redsky-system created
+customresourcedefinition.apiextensions.k8s.io/experiments.redskyops.dev created
+customresourcedefinition.apiextensions.k8s.io/trials.redskyops.dev created
+clusterrole.rbac.authorization.k8s.io/redsky-manager-role created
+clusterrolebinding.rbac.authorization.k8s.io/redsky-manager-rolebinding created
+deployment.apps/redsky-controller-manager created
+clusterrole.rbac.authorization.k8s.io/redsky-patching-role created
+clusterrolebinding.rbac.authorization.k8s.io/redsky-patching-rolebinding created
+secret/redsky-manager created
```
jgustie: This is too specific and too likely to change rendering the documentation out-of-date.

docs/quickstart.md
```diff
@@ -1,70 +1,163 @@
...
+We'll use the Postgres example [here][postgres-example].
+This example will deploy the Postgres application and configure an experiment to tune the memory and CPU limits for Postgres.
+The controller will schedule trials (Kubernetes jobs) using `pgbench` to generate load against our Postgres instance.
+Each trial will test a different set of parameters provided by the Red Sky Ops ML servers.
+The effectiveness of each trial is gauged by the metrics, in this case we contrast cost versus duration.
 
+Deploy the Postgres application and experiment using the following:
+```sh
+$ kustomize build github.com/redskyops/redskyops-recipes/postgres | \
+  kubectl apply -f -
```
jgustie, line 58: To do line continuation in these Markdown blocks you need to use `>` on the subsequent line. Subjectively, indentation helps.

For example `$ kustomize ... \` is on one line and `>   kubectl ...` is on the next (that's three spaces after the gt, it's tough because this comment itself is Markdown).

jgustie (changes requested):

docs/quickstart.md
bradbeam: Not sure exactly what you're getting at here.

If it's around the change in style around links, I felt changing `[text](url)` to `[text][anchor]` worked better in documentation. The anchor link is at the bottom of the document

jgustie: `[minikube][]` == `[minikube][minikube]`

(and the reference is case-insensitive so `[Foo Bar][]` works with `[foo bar]: ...`)

docs/quickstart.md
bradbeam: Some of the feedback we had gotten back was around verifying that things were actually deployed. 
I'm a bit indifferent on the `redskyctl init --wait`, but I think adding the verification for the controller running should remain. 
Perhaps this is something we can add to the `check` command --  `redskyctl check --install`

docs/quickstart.md
bradbeam: These are inside code blocks, so I don't believe `>` is needed. ( check out the preview, they look appropriate )

I know the shell typically inserts `>` to denote being on a newline, but from a usability perspective it prevents a user from just selecting and copying/pasting the command into the shell.

jgustie: Yes, but these are meant to be like screenshots of a console so if the "$" prompt is there, the ">" continuation prompt should be to. This was all stuff I didn't do well initially, but we should fix it now.

I didn't check in Hugo, but:

```console
$ echo \
>   test
test
```

vs.

```console
$ echo \
  test
test
```

jgustie: vs.

```console
$ echo \
> test
test
```

bradbeam: Adding the `>` inside the code block breaks the ability for people to copy/paste.

If we take out the line breaks, then you need to scroll horizontally to see the entire command which isn't great either. 

Until we add in some js or something to the theme to provide a copy to clipboard option that can strip the extra characters automatically, I'd really like to not include the `>`.

docs/quickstart.md
bradbeam: Doing testing locally, it doesn't seem like we have support for syntax highlighting enabled -- https://github.com/redskyops/redskyops-www/blob/staging/config.toml#L7-L8

Enabling it ( with the sh/console changes ) results in:

image

Which seems a little jarring after looking at the theme for so long, but I think I appreciate it better because it highlights the commands to run.

jgustie: I didn't try very hard with the [styles](https://github.com/redskyops/redskyops-www/blob/staging/themes/carbonrelay/static/css/docs.css).

jgustie: I think recent versions of Hugo changed the Markdown processor and syntax highlighter, so the default theme also changed; last I checked changing the config to have:
```
markup:
  highlight:
    style: bw
```
Produced more "boring" highlights (there are plenty of [stock alternatives](https://xyproto.github.io/splash/docs/) though).

docs/quickstart.md
stephanie-rifai: Yeah I agree we should keep it as Quick Start

stephanie-rifai:
```suggestion
Quick Start
```

docs/quickstart.md
stephanie-rifai: do we already have that set up/people access it? If so that works for me!

stephanie-rifai:
```suggestion
If you encounter any issues or require assistance, please reach out to us via [Slack][rso slack] or [contact us][contact us form].

```

docs/quickstart.md
stephanie-rifai: works for me

stephanie-rifai:
```suggestion
Now that you've got an idea of the basics, visit the [experiment lifecycle][lifecycle] and [experiment concepts][concepts] pages to learn more.
```

docs/quickstart.md
bradbeam: I'll drop them as part of the doc. As a followup ( doesn't have to be crammed in this release ) we should try to summarize the output with a simple `success`.

docs/quickstart.md
bradbeam: Dropping this for now as per https://github.com/redskyops/redskyops-controller/pull/143#discussion_r477528954

jgustie (approved):

PR #144, opened 2020-08-25
Interfaceimpl

bradbeam:
Been meaning to chat about this for a bit, but I'd been thinking about this metrics interface refactor for a bit and wanted to share some thoughts around the implementation. 

I think this is in line with the comment of wanting to clean it up later with an actual type. I think the biggest challenge with the proposed implementation is that it creates cyclical dependencies with `Input`. I think `Input` is a great way initialize the various sources ( via `NewSource` ) but I think the actual creation of the sources should do without it. I tried to leave in some more verbose comments to give an idea of where I'm at. 

Unfortunately, git thinks that the file has been deleted instead of renamed ( for prometheus ), so it looks like a larger changes than it should be.


jgustie:
The thought behind having input as a parameter to the capture function was so the "sources" could be longer lived, e.g. the Prometheus source would have a pool of API instances keyed off the URL or the JSON path source would have a pool of HTTP client instances. This is as opposed to creating them for every metric collection. I would need to think about it more and if that use case even makes sense (i.e. is there even any benefit to long lived metric sources).

I wasn't thrilled with the name `Input` for the struct, but I felt it was too many parameters for the capture function and I couldn't come up with something better.

I also didn't consider sub-packaging the individual sources, I'm not opposed to doing it. I'm pretty sure the behavior to request a delayed re-queue is unique to Prometheus: we could refactor that error into `/internal/metric/prometheus`.

jgustie:
I would like to get the change for the original PR in for 1.6.x since it ties off the URL changes, the source refactoring can be a follow on.

jgustie:
I wonder if something halfway in-between the two approaches makes sense; i.e. the `Capture` function accepts a context, the start/end times, and the (evaluated) queries. The metric name and (resolved) URL are part of the source state (i.e. it is likely that multiple metrics would query the same endpoint, eventually we should take advantage of that).

PR #147, opened 2020-08-31
refactor: Move experiment tests from gomega to testify

bradbeam:


jgustie (approved):
Are there any other Gomega tests left? Can we update the `go.mod` yet?

PR #150, opened 2020-09-01
fix: Clean up experiment/trial parseNames logic

bradbeam:
This change addresses an issue with being unable to delete experiments that end in a -digit.

Previously, this would trigger parsing as if it were a trial name even if we were attempting to
handle an experiment.

Hopefully this also makes the code a bit more readable.

### DIFF
--- a/redskyctl/internal/commands/experiments/experiments.go
+++ b/redskyctl/internal/commands/experiments/experiments.go
@@ -39,6 +39,8 @@ const (
 	typeTrial resourceType = "trial"
 )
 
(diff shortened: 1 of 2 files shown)


jgustie:
HA! I did this same thing last night. My concern with this implementation is that we do not impose restrictions on the names anywhere else so nothing would stop you from creating an experiment.yaml with an experiment named "experiment" which you would then be unable to fetch. I'll push my version also.

bradbeam:
closing in favor of #151

PR #151, opened 2020-09-01
Fix experiment name parsing bug

jgustie:


redskyctl/internal/commands/experiments/experiments.go
```diff
@@ -45,14 +44,18 @@ func validTypes() []string {
 }
 
 // normalizeType returns a consistent value based on a user entered type
-func normalizeType(t string) (resourceType, error) {
+func normalizeType(t string) (resourceType, bool, error) {
```
bradbeam: nit, can we add a comment about what this bool is?

redskyctl/internal/commands/experiments/experiments.go
```diff
@@ -101,47 +104,91 @@ func (n *name) experimentName() experimentsv1alpha1.ExperimentName {
...
+		normalType, plural, err := normalizeType(argType)
+		if err != nil {
+			return nil, err
+		}
+		if defaultType == "" && argName == "" && argNumber == "" {
+			defaultType = string(normalType)
+			if plural {
+				// Suspect.
+				defaultType += "s"
+			}
```
bradbeam: won't this append an unnecessary s?

```
defaultType = ""
argType = experiments
normalType = experiments

defaultType = experiments
defaultType = experimentss
```

jgustie: The "s" is always added to `normalType` after it is cast to a string and a normal type will never have the "s". The fact that it's not clear is a pretty good indicator of code smell though.

jgustie: I reworked this a little so the "s" gets added in `normalizeType`. I tried going further and adding a `trialNumber` so we could defer getting the trial numbers until later, but the ambiguity between an experiment named `exp-1` and the first trial of the `exp` experiment causes problems when we are deciding if we need to do a list or get. Without injecting server validation into the logic, it's going to be difficult to address this problem 100%.

redskyctl/internal/commands/experiments/experiments.go
```diff
@@ -101,47 +105,100 @@ func (n *name) experimentName() experimentsv1alpha1.ExperimentName {
...
+
+		// Normalize the type
+		normalType, pluralType, err := normalizeType(argType)
+		if err != nil {
+			return nil, err
+		}
+
+		// Set the default type if it hasn't been set and no name was supplied
+		if defaultType == "" && argIsTypeOnly {
+			defaultType = pluralType
```
bradbeam, line 137: mind adding a comment to clarify that `pluraltype` isn't always plural? I'm sure there's a better way to phrase it, just worried that we'll come back to this code in a bit and be like why are we defaulting this to the plural type.

redskyctl/internal/commands/experiments/experiments.go
```diff
@@ -101,47 +105,100 @@ func (n *name) experimentName() experimentsv1alpha1.ExperimentName {
...
+			}
 		}
 
-		p := strings.SplitN(n.Name, "/", 2)
-		if len(p) > 1 || t == "" {
-			nt, err := normalizeType(p[0])
+		// Parse the number, if present
+		if argNumber != "" {
+			if n.Type != typeTrial {
+				return nil, fmt.Errorf("%s name cannot include a number: %s", n.Type, arg)
```
bradbeam, line 154: just to make sure I'm on the same page; this isn't a new limitation correct?
`experiment-123` is still a valid experiment name, but `experiment/123` would not be?

redskyctl/internal/commands/experiments/experiments.go
```diff
@@ -101,47 +105,100 @@ func (n *name) experimentName() experimentsv1alpha1.ExperimentName {
...
-		if t == "" {
+		if defaultType == "" {
 			return nil, fmt.Errorf("required resource not specified")
 		}
-		names = append(names, name{Type: t, Number: -1})
+		normalType, _, err := normalizeType(defaultType)
+		if err != nil {
+			return nil, err
+		}
+		return []name{{Type: normalType, Number: -1}}, nil
```
bradbeam: nit, can we drop the return here and do either
```
names = []name{Type: normalType, Number: -1}
```
or
```
names = append(names, name{Type: normalType, Number: -1})
```
and let the function return as normal?

redskyctl/internal/commands/experiments/experiments.go
```diff
@@ -101,47 +105,100 @@ func (n *name) experimentName() experimentsv1alpha1.ExperimentName {
...
+		return defType, p[0], p[1]
+	case 3:
+		// type/name/number
+		if p[2] == "" {
+			p[2] = "-1"
+		}
+		return p[0], p[1], p[2]
+	default:
+		// likely invalid name
+		return p[0], strings.Join(p[1:], "/"), ""
```
bradbeam: Are there any cases off the top of your head where this would be ok?

Should we just fail here instead?

bradbeam (dismissed):
nothing major, just a few nits

bradbeam (approved):

PR #152, opened 2020-09-01
chore: Add ci test for doc generation

bradbeam:
Help catch instances where we(I) forgot to generate docs.


jgustie:
Is this just going to log the changes in the CI logs? Or should there be an `--exit-code` to make it fail if there are differences?

bradbeam:
exit code was intended, will update

jgustie (approved):

PR #153, opened 2020-09-01
BuiltIn Metrics

bradbeam:
Wanted to get stuff out there so we could use this for discussion. 

Currently I'm thinking of introducing a new metric type ( MetricBuiltIn ) to facilitate our ability to include metrics out of the box.

This will eventually expand to:
- include a prometheus deployment
- include built in metrics
  - these are ultimately prometheus metrics, but allow for generating the prometheus server url and inject additional metadata ( label selectors, query ranges, etc )
- incorporate the built in metrics automatically as part of a trial run

The workflow would be something like:
- experiment gets created
- prometheus server and kube state metrics get deployed alongside experiment
- during trial run, prometheus server scrapes cadvisor and kube state metrics
- after trial run, user provided metrics and builtin metrics are evaluated
- profit(?)

### DIFF
--- a/internal/metric/builtin.go
+++ b/internal/metric/builtin.go
@@ -30,9 +30,8 @@ import (
 var BuiltIn = []redskyv1beta1.Metric{cpuUtilizationMetric}
 
 var cpuUtilizationMetric = redskyv1beta1.Metric{
-	Name: "rso cpu utilization",
-	Type: redskyv1beta1.MetricBuiltIn,
-	// TODO figure out how we want to inject labels
+	Name:  "rso cpu utilization",
+	Type:  redskyv1beta1.MetricBuiltIn,
 	Query: cpuUtilizationQuery,
 }
 
--- a/internal/metric/builtin.go
+++ b/internal/metric/builtin.go
@@ -35,7 +35,7 @@ scalar(
     sum(
       sum(kube_pod_container_status_running == 1) by (pod)
       *
-      on (pod) group_left kube_pod_labels{label_component="worker"}
+      on (pod) group_left kube_pod_labels{{rsoTargetLabel .Trial}}
     ) by (pod)
     *
     on (pod) group_right max(container_cpu_usage_seconds_total{container="", image=""}) by (pod)
@@ -45,7 +45,7 @@ scalar(
     sum(
       sum(kube_pod_container_status_running == 1) by (pod)
       *
-      on (pod) group_left kube_pod_labels{label_component="worker"}
+      on (pod) group_left kube_pod_labels{{rsoTargetLabel .Trial}}
     ) by (pod)
     *
     on (pod) group_left sum_over_time(kube_pod_container_resource_limits_cpu_cores[1h:1s])
--- a/internal/metric/builtin.go
+++ b/internal/metric/builtin.go
@@ -23,7 +23,7 @@ import (
 var BuiltIn = []redskyv1beta1.Metric{cpuUtilizationMetric}
 
 var cpuUtilizationMetric = redskyv1beta1.Metric{
-	Name: "cpu utilization",
+	Name: "rso cpu utilization",
 	Type: redskyv1beta1.MetricBuiltIn,
 	// TODO figure out how we want to inject labels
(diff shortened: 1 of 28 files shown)


internal/template/template.go
```diff
@@ -52,6 +53,8 @@ type MetricData struct {
 	Values map[string]int64
 	// List of pods from the trial namespace (only available for "pods" type metrics)
 	Pods *corev1.PodList
+	// The experiment identifier
+	ExperimentNamespacedName types.NamespacedName
```
jgustie: Can we just call this `Experiment`? You can reference it in templates as `.Experiment.Name`, `.Experiment.Namespace` or `.Experiment` (which is the "namespace/name" form).

internal/metric/builtin.go
```diff
@@ -0,0 +1,54 @@
...
+    sum(
+      sum(kube_pod_container_status_running == 1) by (pod)
+      *
+      on (pod) group_left kube_pod_labels{{rsoTargetLabel .Trial}}
+    ) by (pod)
+    *
+    on (pod) group_left sum_over_time(kube_pod_container_resource_limits_cpu_cores[1h:1s])
+  )
+)
+`
```
jgustie, line 54: Why isn't this just a template function or a named template that you can use with the existing Prometheus collector? For example, `query: '{{ cpuUtilization "component=redis,component=web" }}'` would effectively evaluate out to the CPU utilization query for pods labeled with either `component=redis` or `component=web`.

internal/template/template_test.go
```diff
@@ -156,14 +215,19 @@ func TestEngine(t *testing.T) {
...
 			case *redskyv1beta1.PatchTemplate:
-				boutput, err = eng.RenderPatch(tc.input.(*redskyv1beta1.PatchTemplate), tc.trial)
+				boutput, err = eng.RenderPatch(inputType, tc.trial)
 				got = string(boutput)
 			case *redskyv1beta1.HelmValue:
-				got, err = eng.RenderHelmValue(tc.input.(*redskyv1beta1.HelmValue), tc.trial)
+				got, err = eng.RenderHelmValue(inputType, tc.trial)
 			case *redskyv1beta1.Metric:
-				got, _, err = eng.RenderMetricQueries(tc.input.(*redskyv1beta1.Metric), tc.trial, tc.obj)
+				// Not sure of a better way to handle this
```
jgustie, line 222: I think we should probably start by splitting `TestEngine` into `TestEngine_RenderHelmValue`, `TestEngine_RenderMetricQueries`, and `TestEngine_RenderPatch`.

internal/template/template.go
```diff
@@ -144,6 +149,25 @@ func (e *Engine) RenderMetricQueries(metric *redskyv1beta1.Metric, trial *redsky
...
+}
+
+func (e *Engine) RenderAnnotationTarget(metric *redskyv1beta1.Metric, trial *redskyv1beta1.Trial) (string, error) {
+	data := newMetricData(trial, nil)
+
+	labelPairBytes, err := e.render("annotation", metric.Query, data)
+
+	return labelPairBytes.String(), err
+}
+
```
jgustie: How are these getting used? I don't think it necessary to leverage templates for this.

With the existing Prometheus metric collection, just change it so if there is nothing specified (i.e. today no labels, in the future, no URL) we just default the Prometheus URL to what we would be standing up. That is, instead of failing when we see missing configuration we will push the potential failure further down the road (when we try to connect).

jgustie (changes requested):
Looks like there are also comments, in progress printlns etc. that need some clean up.

jgustie:
In general I don't think we should add a new metric type for this, I think we should just make the existing Prometheus type smarter about defaulting to our Prometheus server and the using nested templates (or template functions) to generate the more complex PromQL queries.

internal/template/template.go
bradbeam: This has been dropped. 
With the change to a static name for the service, we can use `.Trial` to get the namespace

bradbeam:
Breaking out a couple of the changes to #158 #159 . Will rebase after those land.

PR #155, opened 2020-09-03
Make helm format

jgustie:
When generating a secret, we allow you to specify `--helm-values` and it will emit the YAML necessary for configuring the Helm chart. This PR just makes "helm" an output format so `--helm-values` is now `-o helm`.

### DIFF
--- a/redskyctl/internal/commands/authorize_cluster/generator.go
+++ b/redskyctl/internal/commands/authorize_cluster/generator.go
@@ -31,6 +31,7 @@ import (
 	"golang.org/x/oauth2"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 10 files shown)


redskyctl/internal/commands/authorize_cluster/generator.go
```diff
@@ -67,7 +66,7 @@ func NewGeneratorCommand(o *GeneratorOptions) *cobra.Command {
 		Long:  "Generate authorization secret for Red Sky Ops",
 
 		Annotations: map[string]string{
-			commander.PrinterAllowedFormats: "json,yaml",
+			commander.PrinterAllowedFormats: "json,yaml,helm",
```
bradbeam, line 69: it'd be nice if we could generate these from what's defined through `commander.SetKubePrinter` to avoid mismatches. not something for now though.

bradbeam (approved):

PR #156, opened 2020-09-09
Update quickstart.md to include baseline

stephanie-rifai:


docs/quickstart.md
```diff
@@ -79,8 +80,24 @@ postgres-example-008   Waiting     cpu=1198, memory=2701
...
-
 After the trial is complete, you will be able to view the parameters and the metrics generated from the trial.
+
+## Add the Current Configuration as a Baseline
+
+Optionally, you can compare Red Sky Ops results to a baseline by running a trial with the parameters that are currently set in the the application manifests.
+While your experiment is still running, use the following command to suggest a trial.
+
+```sh
+redskyctl suggest postgres-example --interactive --labels 'baseline=true'
```
bradbeam: Only nit, can you begin this line with `$ `

docs/quickstart.md
```diff
@@ -79,8 +80,24 @@ postgres-example-008   Waiting     cpu=1198, memory=2701
...
+```
+
+You will be prompted to fill in integer values for each parameter in the experiment file.
+Set memory=2000 and CPU=1000.
```
jgustie: Instead of `--interactive` and telling them what to set, should we just use `-A memory=2000 ...` above?

docs/quickstart.md
stephanie-rifai: ah, yes thank you!

docs/quickstart.md
stephanie-rifai: Yeah I went back and forth on this. I think when you have a lot of parameters, the interactive is a better user experience (that's my assumption anyway). I think I will change the command to assign but give the interactive option as well. Objections?

bradbeam (dismissed):

stephanie-rifai:
[TO DO BEFORE MERGING]: update results image to include baseline image

jgustie (approved):

PR #159, opened 2020-09-11
feat: Add prometheus deployment manifests to setup tools

bradbeam:


config/docker-entrypoint.sh
```diff
@@ -1,19 +1,22 @@
...
+  ;;
+  chart)
+    # Package the Helm chart
     shift && /workspace/chart/build.sh "$@"
     exit $?
-fi
+  ;;
+  prometheus)
+    # Generate prometheus manifests
+    shift && cd /workspace/rso
```
jgustie: This should be `/workspace/prometheus` you can just wipe out the existing service monitor (the endpoints it would have been using are gone, it's dead code).

config/docker-entrypoint.sh
```diff
@@ -31,6 +34,8 @@ fi
 
 # Add trial labels to the resulting manifests so they can be more easily located
 if [ -n "$TRIAL" ]; then
+    # Note, this heredoc block must be indented with tabs
+    # <<- allows for indentation via tabs, if spaces are used it is no good.
```
jgustie, line 46: I'm sensing you got burned by this?

config/docker-entrypoint.sh
```diff
@@ -44,12 +49,29 @@ if [ -n "$TRIAL" ]; then
...
+    cat <<-EOF >"experiment_labels.yaml"
+		apiVersion: konjure.carbonrelay.com/v1beta1
+		kind: LabelTransformer
+		metadata:
+		  name: experiment-labels
+		labels:
+		  "redskyops.dev/experiment": $EXPERIMENT
+		  "redskyops.dev/experiment": experimentResource
```
jgustie: This second (duplicate) label should be `experiment-role`; also you should add it to `well_known_labels.go`.

config/docker-entrypoint.sh
```diff
@@ -44,12 +49,29 @@ if [ -n "$TRIAL" ]; then
...
+fi
+
+
 # Process arguments
 while [ "$#" != "0" ] ; do
     case "$1" in
     create)
         handle () {
-            kubectl create -f -
+            kubectl apply -f -
```
jgustie: This was explicitly `create` for `generate-name` and to force failure on conflict. You should add another case for `apply` (it used to be there but got removed after we weren't using it for anything).

config/rso/kube-state-metrics-deployment.yaml
```diff
@@ -0,0 +1,52 @@
+---
+# Source: prometheus/charts/kube-state-metrics/templates/deployment.yaml
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: kube-state-metrics
+  namespace: default
```
jgustie: We shouldn't be putting things in the default namespace like this: I would have expected the namespace to be omitted from all of the resources (even if the edit at runtime is going to add a namespace transformer).

Also, even though the entry point script should create the right kustomization by scanning for resources, it should still be explicitly included in the directory and should still explicitly list all of the resources).

config/rso/prometheus-server-rbac.yaml
```diff
@@ -0,0 +1,63 @@
...
+#      - ingresses
+#    verbs:
+#      - get
+#      - list
+#      - watch
+#  - nonResourceURLs:
+#      - "/metrics"
+#    verbs:
+#      - get
+#
```
jgustie: Remove these comments, do it in a separate commit if you might want them back.

jgustie (changes requested):

config/docker-entrypoint.sh
bradbeam: yup. saw the indentation looked off, #thereifixedit and it no longer worked.

config/prometheus/prometheus-server-service.yaml
```diff
@@ -0,0 +1,35 @@
+---
+# Source: prometheus/templates/services/server.yaml
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app: prometheus
+  name: prometheus-server
```
jgustie: Can we just name this `prometheus` so when we have URL support we can have a default of `http://prometheus:9090` ?

config/prometheus/prometheus-server-service.yaml
```diff
@@ -0,0 +1,35 @@
...
+spec:
+  ports:
+  - name: http
+    port: 9090
+    protocol: TCP
+    targetPort: 9090
+  selector:
+    app: prometheus
+  sessionAffinity: None
+  type: ClusterIP
```
jgustie: Is this a duplicate service?

config/prometheus/prometheus-server-deployment.yaml
```diff
@@ -0,0 +1,76 @@
...
+          args:
+            - --volume-dir=/etc/config
+            - --webhook-url=http://127.0.0.1:9090/-/reload
+          volumeMounts:
+            - name: config-volume
+              mountPath: /etc/config
+              readOnly: true
+        - name: prometheus-server
+          image: "prom/prometheus:latest"
+          imagePullPolicy: "IfNotPresent"
```
jgustie: I would be a little concerned trusting `latest` here, can we pin this to a version (at least a major, maybe major/minor)?

jgustie (dismissed):

config/prometheus/prometheus-server-service.yaml
bradbeam: Want to avoid `prometheus` since that has a potential for conflict. Will use `rso-prometheus` below and remove this one.

bradbeam: along that vein I should probably change all the basic 'prometheus' names to something more specific. ugh.

jgustie: I like the change of the name prefix for all the other objects, I still don't like it on the service because of the whole sensible defaults thing.

A default of "http://prometheus:9090" is a good default because it is likely to actually hit (our Prometheus or theirs, it's generic); if the service name is "redsky-prometheus" then we either need to make the default URL "http://redsky-prometheus:9090" (only likely to hit our Prometheus) or stick with "http://prometheus:9090" and have a default that needs to be overwritten in the particular use case we established the default behavior for in the first place.

I'm fine to revisit this once we are closer with URL based metric collection.

jgustie (approved):

PR #160, opened 2020-09-11
Support for categorical parameters

jgustie:
This changes the primary type of a parameters assignment from `int64` to `IntOrStr` (which is a 32-bit int, more inline with what is used throughout the Kube API). Additionally, it allows bounds to be specified as an array values to indicate a categorical parameter should be created on the backend.

To ensure round-trip-ability between v1alpha1 and v1beta1, the same compatible change was made to both versions (the size of the values list is set as the max on v1alpha1 just be safe).


api/v1alpha1/experiment_types.go
```diff
@@ -36,9 +36,11 @@ type Parameter struct {
 	// The name of the parameter
 	Name string `json:"name"`
 	// The inclusive minimum value of the parameter
-	Min int64 `json:"min,omitempty"`
+	Min int32 `json:"min,omitempty"`
 	// The inclusive maximum value of the parameter
-	Max int64 `json:"max,omitempty"`
+	Max int32 `json:"max,omitempty"`
+	// Internal use only
+	Values []string `json:"values,omitempty"`
```
bradbeam, line 43: Stinks we need to expose this to handle the conversion. Do we need to support the conversion here since it should only be a new field in v1beta1?

internal/server/server.go
```diff
@@ -152,12 +161,17 @@ func ToClusterTrial(t *redskyv1beta1.Trial, suggestion *redskyapi.TrialAssignmen
...
-		if v, err := a.Value.Int64(); err == nil {
-			t.Spec.Assignments = append(t.Spec.Assignments, redskyv1beta1.Assignment{
-				Name:  a.ParameterName,
-				Value: v,
-			})
+		var v intstr.IntOrString
+		if a.Value.IsString {
+			v = intstr.FromString(a.Value.StrVal)
+		} else {
+			v = intstr.FromInt(int(a.Value.Int64Value()))
```
bradbeam: This seems off -- `int(int64)`.
Do we need to handle values > int32 by capping at `math.Maxint32` to avoid rollover?

Some points I found along the way --

> FromInt creates an IntOrString object with an int32 value. It is your responsibility not to call this method with a value greater than int32
https://godoc.org/k8s.io/apimachinery/pkg/util/intstr#FromInt

> int is a signed integer type that is at least 32 bits in size. It is a distinct type, however, and not an alias for, say, int32.
https://golang.org/pkg/builtin/#int

> The int, uint, and uintptr types are usually 32 bits wide on 32-bit systems and 64 bits wide on 64-bit systems.
https://tour.golang.org/basics/11

internal/template/template.go
```diff
@@ -35,7 +36,7 @@ type PatchData struct {
 	// Trial metadata
 	Trial metav1.ObjectMeta
 	// Trial assignments
-	Values map[string]int64
+	Values map[string]interface{}
```
bradbeam, line 39: Should we use intstr here instead of interface?

redskyapi/experiments/v1alpha1/parameter.go
```diff
@@ -0,0 +1,124 @@
...
+		v = numstr.FromNumber(json.Number(s))
+	case ParameterTypeCategorical:
+		v = numstr.FromString(s)
+	}
+	return &v, nil
+}
+
+// CheckParameterValue validates that the supplied value can be used for a parameter.
+func CheckParameterValue(p *Parameter, v *numstr.NumberOrString) error {
+	if p.Type == ParameterTypeCategorical {
```
bradbeam: Should we add this to the switch statement below?

redskyctl/internal/commands/experiments/get.go
```diff
@@ -276,10 +276,12 @@ func (o *GetOptions) filterAndSortTrials(l *experimentsv1alpha1.TrialList) error
 
 // sortableTrialData slightly modifies the schema of the trial item to make it easier to specify sort orders
 func sortableTrialData(item *experimentsv1alpha1.TrialItem) map[string]interface{} {
-	assignments := make(map[string]int64, len(item.Assignments))
+	assignments := make(map[string]interface{}, len(item.Assignments))
```
bradbeam, line 279: Should we use numstr here instead?

redskyctl/internal/commands/experiments/suggest.go
```diff
@@ -181,66 +177,71 @@ func (o *SuggestOptions) assignInteractive(p *experimentsv1alpha1.Parameter, def
...
+func randomValue(p *experimentsv1alpha1.Parameter) (*numstr.NumberOrString, error) {
 	switch p.Type {
 	case experimentsv1alpha1.ParameterTypeInteger:
-		min, max, err := intBounds(&p.Bounds)
+		min, max, err := intBounds(p.Bounds)
 		if err != nil {
 			return nil, err
 		}
-		r := json.Number(strconv.FormatInt(rand.Int63n(max-min)+min, 10))
+		r := numstr.FromInt64(rand.Int63n(max-min) + min)
```
bradbeam, line 229: not related to this pr, but do we seed the rand here?

api/v1alpha1/experiment_types.go
jgustie: Just going to quote the Kubebuilder book:
> In Kubernetes, all versions must be safely round-tripable through each other. This means that if we convert from version 1 to version 2, and then back to version 1, we must not lose information. Thus, any change we make to our API must be compatible with whatever we supported in v1, and also need to make sure anything we add in v2 is supported in v1. In some cases, this means we need to add new fields to v1, but in our case, we won’t have to, since we’re not adding new functionality.

internal/template/template.go
jgustie: We cannot or we would break the template functions, we actually unroll the values before we put them in the map so the appropriate type is available reflectively in the templates.

redskyctl/internal/commands/experiments/suggest.go
jgustie: We seed the global rand cryptographically in the `init()` of the main package.

redskyapi/experiments/v1alpha1/parameter.go
jgustie: This if statement handles the string case (we can't make it down to the switch, the `IsString` check would fail), the switch below is for the numeric checks. This if block ends with a return.

redskyctl/internal/commands/experiments/get.go
jgustie: We need to pass JSON primitives down into the sorting code.

internal/server/server.go
jgustie: I went back and forth on this. At the end of the day the bounds we collect in the Kube API are strongly typed to `int32` so if the server honors the configured bounds then we will never overflow. Everything in `suggest` is done with `int64` math because we do not technically have the protection of what is in the Kube API when dealing directly with the ML backend (in fact we still deal with floating point numbers there). The `FromInt` call will report overflow on 64-bit architectures to the log.

I will update this, but because we aren't reporting failures, the behavior will likely be to just cap values to what fits in a 32-bit integer: again the assumption being that the code would be effectively unreachable.

jgustie:
Does the UI work with string values?

bradbeam (approved):

PR #162, opened 2020-09-11
Refactor: Remove additional gomega tests

bradbeam:


jgustie (approved):
I'm not a fan of the authorizationcode_test changes; just replacing `g.Expect` with `assert.Equals` would have been enough. The old test used the `Test<Type>_<Func>` convention for name since that was what all that was being tested. There wasn't multiple test cases so there was no need for the loop. If we are going to test `ExchangeWithPKCE` (using a test server like you suggest, we would do that in a separate `TestConfig_ExchangeWithPKCE` fixture.

PR #163, opened 2020-09-14
Update suggest to send labels via `CreateTrial`

jgustie:
It turns out you can only apply labels to completed trials, since manually created trials will be "staged" they cannot be labeled using the `LabelTrial`. This PR adds a label map to the assignments so they can be sent up when the trial is created.


jgustie:
I just verified this is working in `john-dev`.

chrislarson1 (approved):

PR #165, opened 2020-09-16
feat: Add cpu utilization template function

bradbeam:
This introduces the cpuUtilization template function which evaluates to
a prometheus query to grab cpu utilization metrics for a trial.

### DIFF
--- a/internal/template/function.go
+++ b/internal/template/function.go
@@ -85,20 +85,20 @@ func resourceRequests(pods corev1.PodList, weights string) (float64, error) {
 	return totalResources, nil
 }
 
-func cpuUtilization(labelArgs ...string) (string, error) {
+func cpuUtilization(data MetricData, labelArgs ...string) (string, error) {
 
 	cpuUtilizationQueryTemplate := `
 scalar(
   sum(
-    increase(container_cpu_usage_seconds_total{container="", image=""}[1h]) by (pod)
+    increase(container_cpu_usage_seconds_total{container="", image=""}[{{ .Range }}]) by (pod)
     *
-    on (pod) group_left kube_pod_labels{{ . }}
+    on (pod) group_left kube_pod_labels{{ .Labels }}
   )
   /
   sum(
-    sum_over_time(kube_pod_container_resource_limits_cpu_cores[1h:1s])
+    sum_over_time(kube_pod_container_resource_limits_cpu_cores[{{ .Range }}:1s])
     *
-    on (pod) group_left kube_pod_labels{{ . }}
+    on (pod) group_left kube_pod_labels{{ .Labels }}
   )
 )`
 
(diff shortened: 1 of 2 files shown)


internal/template/function.go
```diff
@@ -82,3 +84,50 @@ func resourceRequests(pods corev1.PodList, weights string) (float64, error) {
...
+    sum_over_time(kube_pod_container_resource_limits_cpu_cores[1h:1s])
+    *
+    on (pod) group_left kube_pod_labels{{ . }}
+  )
+)`
+
+	tmpl, err := template.New("query").Parse(cpuUtilizationQueryTemplate)
+	if err != nil {
+		return "", err
+	}
```
jgustie: `tmpl := template.Must(template.New("query").Parse(cpuUtilizationQueryTemplate))`

internal/template/function.go
```diff
@@ -82,3 +84,50 @@ func resourceRequests(pods corev1.PodList, weights string) (float64, error) {
...
+
+	var labels []string
+	for _, label := range strings.Split(strings.Join(labelArgs, ","), ",") {
+		kvpair := strings.Split(label, "=")
+		if len(kvpair) != 2 {
+			// Should we continue or hard fail on invalid labels?
```
jdplatt: I would fail the metric scrape if the labels are invalid

internal/template/function.go
```diff
@@ -82,3 +84,50 @@ func resourceRequests(pods corev1.PodList, weights string) (float64, error) {
...
+			continue
+		}
+
+		labels = append(labels, fmt.Sprintf("label_%s=\"%s\"", kvpair[0], kvpair[1]))
+	}
+
+	var labelParam string
+	// if 0 labels, should we inject namespace=.Trial.Namespace automatically?
```
jgustie: I do not believe we have this context available with explicitly accepting the trial as an argument do we?

internal/template/function.go
```diff
@@ -82,3 +84,50 @@ func resourceRequests(pods corev1.PodList, weights string) (float64, error) {
 	}
 	return totalResources, nil
 }
+
+func cpuUtilization(labelArgs ...string) (string, error) {
+
+	cpuUtilizationQueryTemplate := `
+scalar(
+  sum(
+    max(container_cpu_usage_seconds_total{container="", image=""}) by (pod)
```
jdplatt: I think this should be increase over the time range not max because pods that aren't patched won't start from container_cpu_usage_seconds_total=0

internal/template/function.go
bradbeam: A panic here felt too heavy.
.. granted, it's an internal packate/template so this shouldn't ever fail during normal run conditions.

jgustie: This is exactly when you _should_ panic: if the code itself is invalid (the constant doesn't parse) and the only action is to fix the code.

bradbeam: Sorry, meant more in line with when this gets called it feels too late in the lifecycle.

I think more appropriately we should call the `template.Must` in an init function so we could panic early and reuse the template instead of creating a new one each time. But that felt like a larger change/discussion since we have other instances of template.New being called inline and I didn't want to distract from this. 

The above are more concerns/refactors for later, I'm fine with changing to template.must inline here.

internal/template/function.go
bradbeam: ah we do not

jgustie (approved):

PR #173, opened 2020-09-18
feat: Add avgUtilization function

bradbeam:
This adds a helper function to average results from memory and cpu utilization into a single metric.

### DIFF
--- a/internal/template/utilization_functions.go
+++ b/internal/template/utilization_functions.go
@@ -23,9 +23,8 @@ import (
 	"text/template"
 )
 
-func cpuUtilization(data MetricData, labelArgs ...string) (string, error) {
-
-	cpuUtilizationQueryTemplate := `
+var (
+	cpuUtilizationQueryTemplate = `
 scalar(
   round(
     (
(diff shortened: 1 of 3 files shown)


internal/template/utilization_functions.go
```diff
@@ -68,10 +63,21 @@ scalar(
...
+func cpuUtilization(data MetricData, labelArgs ...string) (string, error) {
+	return renderUtilization(cpuUtilizationQueryTemplate, data, labelArgs...)
+}
+
+func memoryUtilization(data MetricData, labelArgs ...string) (string, error) {
 	return renderUtilization(memoryUtilizationQueryTemplate, data, labelArgs...)
 }
 
+func avgUtilization(data MetricData, labelArgs ...string) (string, error) {
+	avgQuery := fmt.Sprintf("\n(%s\n+%s\n)\n/2", cpuUtilizationQueryTemplate, memoryUtilizationQueryTemplate)
```
jgustie, line 77: Just to be clear, without this it's just `query: '({{cpuUtilization . foo=bar}}+{{memoryUtilization . foo=bar}})/2'` right?

bradbeam: I haven't tested doing both independently like that, but yeah that's what it boils down to.

PR #179, opened 2020-09-23
feat: Add template functions for cpu and memory requests

bradbeam:
This introduces two new template functions, `cpuRequests` and `memoryRequests`, so we can calculate resource usage for the duration of the trial.

### DIFF
--- a/internal/template/utilization_functions.go
+++ b/internal/template/utilization_functions.go
@@ -24,7 +24,6 @@ import (
 )
 
 func cpuUtilization(data MetricData, labelArgs ...string) (string, error) {
-
 	cpuUtilizationQueryTemplate := `
 scalar(
   round(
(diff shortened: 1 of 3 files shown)


internal/template/utilization_functions.go
```diff
@@ -72,6 +71,39 @@ scalar(
...
+    avg_over_time(kube_pod_container_resource_requests_memory_bytes[{{ .Range }}])
+    *
+    on (pod) group_left kube_pod_labels{{ .Labels }}
+  )
+  /
+  1024
+  /
+  1024
+  /
+  1024
```
jgustie: I think this is the only bit I'm questioning, it seems like it would be easier and safer to just leave this in `B` instead of `GiB` (not `MB` like the comment). I'm pretty sure it would be safe to do something like:
```
query: '{{ memoryRequests . "foo=bar" }} / 1024 / 1024 / 1024'
```

jgustie (approved):

PR #182, opened 2020-09-30
Ready condition for status fields

jgustie:
This adds support for a `redskyops.dev/status-<FIELD>-<VALUE>` condition type in readiness checks. The intent is that this can still work with Kubernetes objects like pods, e.g. `redskyops.dev/status-phase-running` would be "not ready" until the pod's `{.status.phase} == "running"`. This can also be useful for CRDs which do not have a `{.status.conditions}` array (which until recently included our own experiments object).

### DIFF
--- a/internal/ready/ready.go
+++ b/internal/ready/ready.go
@@ -168,20 +168,23 @@ func (r *ReadinessChecker) unstructuredConditionStatus(obj *unstructured.Unstruc
 func (r *ReadinessChecker) statusField(obj *unstructured.Unstructured, conditionType string) (string, corev1.ConditionStatus, error) {
 	// In this case the condition type is "redskyops.dev/status-<FIELD>-<VALUE>" so we must parse out the field and value
 	kv := strings.SplitN(strings.TrimPrefix(conditionType, ConditionTypeStatus), "-", 2)
+	if len(kv) != 2 {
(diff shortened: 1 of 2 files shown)


internal/ready/ready.go
```diff
@@ -150,10 +159,37 @@ func (r *ReadinessChecker) unstructuredConditionStatus(obj *unstructured.Unstruc
...
+	}
+
+	v, ok := s[kv[0]].(string)
+	if !ok {
+		// This is a legitimate "unknown" case because we didn't see the field
+		return "", corev1.ConditionUnknown, nil
+	}
+
+	// Check the value case-insensitively
+	if len(kv) > 1 && strings.EqualFold(kv[1], v) {
```
jgustie: I wasn't sure if `len(kv) != 2` should just be an error or not.

bradbeam: Yeah, I think it'd be worth adding some additional validation on this to ensure we have a `status-field-value` pattern. Probably after the split, `if len(kv) != 2` then fail/error.

bradbeam (dismissed):

jgustie:
I rebased this over to a release branch cut from 1.6.4.

bradbeam (approved):

PR #185, opened 2020-10-08
Refactor: generate install

jgustie:
This PR is largely a refactoring of the `redskyctl init` and `redskyctl generate install` commands: it eliminates the duplicate logic that existing in both places by pushing all of the generation to generator.

The main reason for this change is that we are increasingly seeing a preference to get the manifests rather then doing an init directly into the cluster; currently you must run three commands to collect all the manifests (and they all have inter-related configuration). With all of the manifest generation in the generator, the `redskyctl generate install` command now produces the same manifests you get with `redskyctl init`: you do not need to run multiple commands.

Additionally the `/config/install` directory has been removed as it is not longer necessary.


redskyctl/internal/commands/initialize/filters.go
```diff
@@ -0,0 +1,155 @@
...
+
+			if err := commentOutEnvFrom(n); err != nil {
+				return nil, err
+			}
+		}
+		return nodes, nil
+	})
+}
+
+func commentOutEnvFrom(node *yaml.RNode) error {
```
jgustie: This was a pain, but I felt it produced a much more desirable outcome: instead of completely removing the reference to the secret, it is commented out. This is primarily for doing `redskyctl generate install --skip-secret` followed by `redskyctl generate secret` (once you generate the secret, you need to manually uncomment the reference to it). The alternative was to add an `optional: true` when using `--skip-secret` (like we do with the Helm chart), however that seem to be less reliable (and technically, `--skip-secret` _is_ undocumented).

bradbeam: This seems a bit heavy. 
Would it be more appropriate to always include `envFrom.secret` and then use the `skip-secret` to control generating a secret with/without content.

redskyctl/internal/commands/initialize/initialize.go
```diff
@@ -127,103 +84,14 @@ func (o *Options) initialize(ctx context.Context) error {
...
-		kustomize.WithImage(o.Image),
-		kustomize.WithLabels(map[string]string{
-			"app.kubernetes.io/version":    version.GetInfo().Version,
-			"app.kubernetes.io/managed-by": "redskyctl",
-		}),
-		kustomize.WithAPI(apiEnabled),
-	)
-
-	if err != nil {
+	opts := o.GeneratorOptions // This MUST be a copy
```
bradbeam: Mind expanding on why this must be a copy ( is it for iostreams interaction? )

redskyctl/internal/commands/initialize/filters.go
jgustie: Fair enough, my concern is that actually including the `envFrom` will then conflict with the patch generated by `redskyctl authorize-cluster` to add the secret reference to the deployment later.

I tried to clean up the commenting code a bit, but maybe I also need to fix the patch on `authorize-cluster` so we can do an empty secret.

jgustie: Actually, it looks like there is no merge key on `envFrom` so the `authorize-cluster` smp does the right thing (relatively speaking, there is an unrelated bug with metadata though: PR incoming on that front)

bradbeam (approved):

PR #188, opened 2020-10-09
Sorting is broken

jgustie:
I'm not even sure this fixes it.


jgustie:
I added some tests, I'm a little more confident that this does in fact make sorting better. It very quietly falls back to a stable order (`i < j`) when it can't compare things. The server appears to sort by status, then number: so at the very least, this fix makes it so `--sort-by number` behaves as expected.

bradbeam (approved):

PR #190, opened 2020-10-09
feat: Add default label selector for prometheus

bradbeam:
If no label selector is given for prometheus metrics, we'll default to `app: prometheus`.

### DIFF
--- a/controllers/metric_controller.go
+++ b/controllers/metric_controller.go
@@ -202,11 +202,15 @@ func (r *MetricReconciler) target(ctx context.Context, namespace string, m *reds
 	case redskyv1beta1.MetricPods:
 		// Use the selector to get a list of pods
 		target := &corev1.PodList{}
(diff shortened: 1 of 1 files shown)


controllers/metric_controller.go
```diff
@@ -216,11 +220,23 @@ func (r *MetricReconciler) target(ctx context.Context, namespace string, m *reds
 			return target, nil
 		}
 
-		if sel, err := meta.MatchingSelector(m.Selector); err != nil {
+		// Default label selector to `app: prometheus` if not specified
+		labelSelector := m.Selector
+		if labelSelector == nil {
+			labelSelector = &metav1.LabelSelector{
+				MatchLabels: map[string]string{"app": "prometheus"},
```
jgustie, line 227: This doesn't apply to JSON path metrics.

jgustie (approved):

PR #193, opened 2020-10-09
Allow generate install to write files to a directory

jgustie:


bradbeam (approved):

jgustie:
This was one of the benefits of moving the kyaml pipeline into the generator. We can reorganize the files anyway we want by applying annotations, the default layout is `[{namespace}/]{kind}_{name}.yaml`.

PR #194, opened 2020-10-09
fix: Sanitize prometheus query labels

bradbeam:
Prometheus translates kubernetes labels to fit in the alphanumeric and underscore
character class. We should handle cases where the user is not expecting this.


jgustie:
image

jgustie (approved):

PR #138, opened 2020-08-18
Add support for patch command

bradbeam:
Let's try this again. 

This should be a bit more concise and probably closer in line with expectations.

Example usage
```shell 
bradbeam@cr-2:~/go/src/github.com/redskyops/redskyops-controller/redskyctl
[22:36:01] # kustomize build $recipes/postgres | ./redskyctl patch --file - --trialnumber 24 | grep -e cpu -e mem
            cpu: 615m
            memory: 500Mi
            cpu: 615m
            memory: 500Mi
bradbeam@cr-2:~/go/src/github.com/redskyops/redskyops-controller/redskyctl
[22:36:13] # kustomize build $recipes/postgres | ./redskyctl patch --file - --trialnumber 25 | grep -e cpu -e mem
            cpu: 585m
            memory: 1275Mi
            cpu: 585m
            memory: 1275Mi
bradbeam@cr-2:~/go/src/github.com/redskyops/redskyops-controller/redskyctl
[22:36:17] # ./redskyctl patch --file $recipes/postgres/experiment.yaml --file $recipes/postgres/application/postgres.yaml --trialnumber 24 | grep -e cpu -e mem
            cpu: 615m
            memory: 500Mi
            cpu: 615m
            memory: 500Mi
bradbeam@cr-2:~/go/src/github.com/redskyops/redskyops-controller/redskyctl
[22:36:34] # ./redskyctl patch --file $recipes/postgres/experiment.yaml --file $recipes/postgres/application/postgres.yaml --trialnumber 25 | grep -e cpu -e mem
            cpu: 585m
            memory: 1275Mi
            cpu: 585m
            memory: 1275Mi
```

replaces #110 #120 

Signed-off-by: Brad Beam <brad.beam@carbonrelay.com>

PR #199, opened 2020-10-20
Refactor experiment reader

jgustie:
I know this conflicts with #138 but there are a couple of difference here. Most importantly, it's no longer "readExperiment" because nothing in the logic tied things to experiments. This fixes the awkward "I just want to read a single experiment" problem we had before because now you can just choose to read a single experiment instead of a list.

There is also a de-coupling of the file IO from the reading: e.g. the filename itself is no longer passed down (however we do try to recover the `os.File` to check for the filename extension). This makes it easier to pass in any reader without using "-" as a filename programmatically.

Finally any defaults registered with the scheme are applied and the target's type metadata is populated for verification.

My plan is to also use this for reading in the new "Application" type in #184.

### DIFF
--- a/redskyctl/internal/commander/reader.go
+++ b/redskyctl/internal/commander/reader.go
@@ -49,7 +49,7 @@ func NewResourceReader() *ResourceReader {
 	_ = redskyv1beta1.AddToScheme(rr.Scheme)
 	_ = redskyv1alpha1.AddToScheme(rr.Scheme)
 
-	// Allow single experiments to fulfil target an experiment list
+	// Allow single experiments to target an experiment list
 	_ = addExperimentListConversions(rr.Scheme)
 
 	return rr
@@ -104,11 +104,13 @@ func (r *ResourceReader) objectKind(obj runtime.Object) (schema.GroupVersionKind
 	if err != nil {
 		return schema.GroupVersionKind{}, err
 	}
+
 	gvk, ok := r.PreferredVersion.KindForGroupVersionKinds(gvks)
 	if !ok {
 		// Your code must supply a GroupVersioner to disambiguate this case
 		panic("read destination type is ambiguous in scheme")
 	}
+
 	return gvk, nil
 }
 
@@ -119,6 +121,7 @@ func (r *ResourceReader) decoder(mediaType string) (runtime.Decoder, error) {
 	if !ok {
 		return nil, fmt.Errorf("could not find serializer for %s", mediaType)
 	}
+
 	return cs.DecoderToVersion(info.Serializer, r.PreferredVersion), nil
 }
 
@@ -132,6 +135,7 @@ func mediaType(r io.ReadCloser) string {
 			mt = runtime.ContentTypeJSON
 		}
 	}
+
 	return mt
 }
 
@@ -145,6 +149,7 @@ func (onlyVersion) KindForGroupVersionKinds(kinds []schema.GroupVersionKind) (sc
 	if len(kinds) == 1 {
 		return kinds[0], true
 	}
+
 	return schema.GroupVersionKind{}, false
 }
 
@@ -167,5 +172,6 @@ func addExperimentListConversions(s *runtime.Scheme) error {
 	}); err != nil {
 		return err
 	}
+
 	return nil
 }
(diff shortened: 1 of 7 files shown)


redskyctl/internal/commander/reader.go
```diff
@@ -0,0 +1,177 @@
...
+
+	// Allow single experiments to target an experiment list
+	_ = addExperimentListConversions(rr.Scheme)
+
+	return rr
+}
+
+// ReadInto decodes the supplied byte stream into the target runtime object. The default values
+// and type information of the target object will be populated.
+func (r *ResourceReader) ReadInto(reader io.ReadCloser, target runtime.Object) error {
```
bradbeam, line 60: I think only issue I have is `ReadInto(runtime.Object)` seems too generic when we only support our objects. 
Should we extend the reader to support all kube objects or rename `ReadInto` to something more specific ( `ReadRedSkyObject` or something like that ).

redskyctl/internal/commander/reader.go
```diff
@@ -0,0 +1,177 @@
...
+		case "json":
+			mt = runtime.ContentTypeJSON
+		}
+	}
+
+	return mt
+}
+
+// OnlyVersion is a group version that only resolves if there is a single possible kind
+var OnlyVersion runtime.GroupVersioner = onlyVersion{}
```
bradbeam, line 143: Can you expand on the usefulness of this?

redskyctl/internal/commander/reader.go
```diff
@@ -0,0 +1,177 @@
...
+
+	return cs.DecoderToVersion(info.Serializer, r.PreferredVersion), nil
+}
+
+// mediaType returns the presumed media type for the supplied read closer.
+func mediaType(r io.ReadCloser) string {
+	// For now just assume YAML unless we got a JSON file
+	mt := runtime.ContentTypeYAML
+	if f, ok := r.(*os.File); ok {
+		switch filepath.Ext(f.Name()) {
```
bradbeam, line 133: This always has bothered me for flexing on file extension. Should we try something similar to this instead?

> // YAMLOrJSONDecoder attempts to decode a stream of JSON documents or
> // YAML documents by sniffing for a leading { character.

https://github.com/kubernetes/apimachinery/blob/d4f471b82f0a17cda946aeba446770563f92114d/pkg/util/yaml/decoder.go#L326-L348

bradbeam (approved):

redskyctl/internal/commander/reader.go
jgustie: We only support our objects by default to cut down on boilerplate, nothing stops you from doing a `_ = clientgoscheme.AddToScheme(rr.Scheme)` and then using that `ResourceReader` to read a core Kube object.

redskyctl/internal/commander/reader.go
jgustie: A group versioner is used to disambiguate the case when you register the same Go type under multiple GVKs. There are two implementations provider: disabled and internal. Disabled just returns false (which, now that I think about might be good enough if it only gets called if the `len(gvks) > 1`). The internal implementation looks for a special group name and then falls back to the first registered GVK (for some unknown definition of "first").

This group versioner basically says "if you registered the same Go object into the scheme twice, you need to bring your own `GroupVersioner` or else we panic.

redskyctl/internal/commander/reader.go
jgustie: I don't believe it is necessary to that, this is only a hint as to which parser to use. Technically JSON _is_ YAML so even if we miss the extension (e.g. `-f <(cat foo.json)`) we still end up doing the right thing (even though we don't have YAML tags on our struct fields).

I don't know what a stream of JSON documents looks like (ndjson?), but as of right now I am not using the streaming decoders (which use `ErrShortBuffer` to indicate document boundaries). We can consider adding that in later (I left a TODO), but I think it's a lot of work (e.g. we need to validate the target is a list and all documents have a valid type to go into that list). If that logic all already exists somewhere, I would be glad to use it, but that didn't fit in this timebox.

redskyctl/internal/commander/reader.go
jgustie: It looks like the existing "disabled" implementation won't quite work.

PR #200, opened 2020-10-21
Add a ping command

jgustie:
Saw this somewhere yesterday and realized that it makes much more sense then what we currently have. "Ping" is a pretty well understood "can I talk to the server?" command. While technically this functionality already exists (it is part of check when doing `redskyctl check config`), it's far less obvious.

The `redskyctl check server` command is a hold over from when we were first building the controller: I do not think it is necessary anymore, we should move similar functionality out to an automated test rather then exposing it in the tool. Until we decide where to move it, I have just hidden it.


bradbeam (dismissed):

bradbeam:
guessing it needs to rebase off cobra changes?

bradbeam (approved):

PR #203, opened 2020-10-21
Upgrade Cobra

jgustie:
Taking the latest version of Cobra, as they finally addressed the `Command.Print` issues that were preventing me from eliminating the `ExitOnError` code.

I also took the chance to normalize the command short descriptions and flag names to match the Cobra defaults: and put a test in place to enforce it.

This looks like a huge review, but it's mostly just the fact that I previously thought I should make flag usages into sentences which doesn't match the default usage introduced by Cobra.


redskyctl/main.go
```diff
@@ -38,6 +41,9 @@ func init() {
 		panic(err)
 	}
 	rand.Seed(int64(binary.LittleEndian.Uint64(b[:])))
+
+	// Prevent Cobra from changing the command order
+	cobra.EnableCommandSorting = false
```
bradbeam, line 46: curious why?

redskyctl/internal/commander/commander.go
```diff
@@ -173,55 +171,28 @@ func AddPreRunE(cmd *cobra.Command, preRunE func(*cobra.Command, []string) error
...
-// ExitOnError converts all the error returning run functions to non-error implementations that immediately exit
-func ExitOnError(cmd *cobra.Command) {
-	// Convert a RunE to a Run
-	wrapE := func(runE func(*cobra.Command, []string) error) func(*cobra.Command, []string) {
-		return func(cmd *cobra.Command, args []string) {
-			err := runE(cmd, args)
-			if err == nil {
-				return
+// MapErrors wraps all of the error returning functions on the supplied command (and it's sub-commands) so that
+// pass any errors through the mapping function.
```
bradbeam: nit, comment wording is a little broken, maybe missing `they` at the end of L174?

bradbeam (dismissed):

redskyctl/main.go
jgustie: The alphabetical ordering was meaningless, the current ordering is a little more functional/user oriented. Still not quite what I would like, but there is no easy "command groupings" that I am aware of (I think Kube had to build that themselves).

bradbeam (approved):

PR #204, opened 2020-10-22
Improve parameter docs; document categorical params, constraints

jameshochadel:
Going to leave my own review with a few questions for you all.

### DIFF
--- a/docs/parameters.md
+++ b/docs/parameters.md
@@ -1,4 +1,4 @@
-# Using Parameters
+# Parameters
 
 A **parameter** is a {value, field, placeholder, input to an experiment} which can be adjusted during an experiment. Every parameter has a name and a **domain** of acceptable values it can be **assigned** during a trial. 
 
@@ -27,7 +27,7 @@ While a parameter for tuning the concurrent garbage collection threads on a Java
     max: 8
 ```
 
-*Note:* Some fields in Kubernetes objects, like [CPU Request](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#meaning-of-cpu), allow their value to be specified using decimal numbers like `0.1` or as an integer with an attached unit, like `100m` ("one hundred millicpus"). Because Red Sky Ops only tunes integer numbers, you must use the latter format, as explained further in **Using Parameters**. 
+*Note:* Some fields in Kubernetes objects, like [CPU Request](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#meaning-of-cpu), allow their value to be specified using decimal numbers like `0.1` or as an integer with an attached unit, like `100m` ("one hundred millicpus"). Because Red Sky Ops only tunes integer numbers, you must use the latter format, as explained further in [Using Parameters](#using-parameters). 
 
 ### Categorical
 
@@ -47,7 +47,7 @@ Categorical parameters specify a finite list of acceptable values, where each va
 
 The `parameters` section of an experiment file defines the names and domains of the experiment's parameters. The `patches` section defines how those parameters are used in your application.
 
-To reuse an earlier example, an integer parameter for tuning CPU on a container might be defined as follows:
+To continue with an earlier example, an integer parameter for tuning CPU on a container might be defined as follows:
 
 ```yaml
   parameters:
@@ -80,4 +80,4 @@ Sometimes it is necessary to manipulate a parameter value to consume it in a pat
 - **percent**
   Return the integer percentage.
 
-  `percent 9 50` will return `"4"`
+  `percent 9 50` will return `"4"`.
--- a/docs/parameters.md
+++ b/docs/parameters.md
@@ -1,6 +1,6 @@
 # Parameters
 
-A **parameter** is a {value, field, placeholder, input to an experiment} which can be adjusted during an experiment. Every parameter has a name and a **domain** of acceptable values it can be **assigned** during a trial. 
+A **parameter** is a value which can be adjusted during an experiment. Every parameter has a name and a **domain** of acceptable values it can be **assigned** during a trial.
 
 An experiment must define one or more parameters. Together, the parameters in an experiment define the **search space**.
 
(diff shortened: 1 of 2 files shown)


docs/parameters.md
```diff
@@ -1,10 +1,53 @@
-# Using Parameters
+# Parameters
 
-Experiment parameters define the search space for assigned values that vary for each trial run. Each parameter represents a named integer assignment with an inclusive minimum and maximum bound.
+A **parameter** is a value which can be adjusted during an experiment. Every parameter has a name and a **domain** of acceptable values it can be **assigned** during a trial.
```
jameshochadel: > A **parameter** is a value which

Other things I considered besides "value": "field", "placeholder", "input to an experiment". I'm open to other options.

docs/parameters.md
```diff
@@ -1,10 +1,53 @@
...
 
-## Parameter Domain
+An experiment must define one or more parameters. Together, the parameters in an experiment define the **search space**.
 
-When selecting the bounds for a parameter it is important to remember that all values are configured as integers. When tuning compute resources, such as a [CPU request](https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/#meaning-of-cpu), you may typically use values like 0.1, 0.2, 0.3, ..., 4.0. However, for optimization you will need to specify your bounds using millicpus and add the explicit unit later, for example:
+## Parameter Types
+
+Red Sky Ops supports two types of parameters: **Integer** and **Categorical**.
```
jameshochadel: Is it preferable to use the RSO product name, or `redskyops-controller`, or something else?

docs/parameters.md
```diff
@@ -13,18 +56,61 @@ When selecting the bounds for a parameter it is important to remember that all v
...
+
+## Parameter Constraints
+
+**Constraints** restrict the assignments that parameters may take relative to one another. Constraints are optional and used in addition to the bounds on individual parameters like `min` or `max`.
+
+Red Sky Ops supports two types of constraints: **Order** and **Sum**.
+
+### Order Constraint
+
+The order constraint requires that one integer parameter be strictly larger than another. For example, you can use an order constraint to ensure that a container's maximum replicas value is always greater than it's minimum replicas:
```
jameshochadel, line 93: Do order constraints only apply to integers, or do we do any kind of funky string comparison? Also, is it accurate to say "strictly larger"?

docs/parameters.md
```diff
@@ -13,18 +56,61 @@ When selecting the bounds for a parameter it is important to remember that all v
...
+  constraints:
+    - sum:
+        bound: 4000
+        isUpperBound: true
+        parameters:
+          - name: first_cpu
+          - name: second_cpu
+```
+
+Each parameter in the sum constraint may have an associated **weight**. When a weight is specified for a parameter, the parameter's value is multiplied by that weight before being summed.
```
jameshochadel, line 116: If anyone has a good example usage of weighted sum constraints, I'll add it here.

docs/parameters.md
stephanie-rifai: I personally like the idea of variable vs. outcome or input vs. output, where variable/output=metrics

docs/parameters.md
```diff
@@ -1,10 +1,53 @@
-# Using Parameters
+# Parameters
```
jgustie, line 1: This first line needs to be updated in `redskyops-www/build.sh` as well. You should also update the `/docs/metrics.md` to be consistent.

docs/parameters.md
jgustie: Why so much emphasis?

docs/parameters.md
stephanie-rifai: I think red sky ops is fine

jgustie: You do your best to be agnostic when identifying the product ("Parameters can be one of two types..."). Names change and it makes much harder to keep the documentation up to date when it's riddled with specific names.

That being said, I don't know why we are exposing "parameter types" on the user facing documentation. A parameter either has a numeric range (you specify min/max) or a string range (you specify the allowed values). It feels like this is making things more complex than they are. Maybe it's just me.

docs/parameters.md
jameshochadel: @stephanie-rifai Those are helpful, thanks. Think I might switch to input (Though I think you meant *output*/outcome=metrics?)

@jgustie I added emphasis each time a new term was introduced to improve readability. Though it's over-the-top in this sentence because of the emphasis on `assigned`, which a) makes three emphasized terms in close proximity, which is a lot, and b) doesn't look like a technical term. I'll split it out into another sentence.

docs/parameters.md
jameshochadel: > You do your best to be agnostic when identifying the product

My favorite solution: Avoiding the problem 😄 

---
Re. "parameter types": Honestly, I mentally organized them this way so automatically that I didn't even consider whether it was appropriate. I do think it's a helpful distinction though:

* It's consistent with how we talk about them ("We added categorical parameters in X release")
* The syntax for defining their ranges is distinct
* Certain features — parameter constraints for instance — only apply to integer parameters
* Integer ranges have certain intrinsic properties that string ranges won't (values are more meaningfully comparable, ranges are defined without explicitly defining every value)

That said, I could certainly de-emphasize the difference if you think it's best.

docs/parameters.md
stephanie-rifai: yes, sorry!

docs/parameters.md
jameshochadel: Will do, thanks

docs/parameters.md
stephanie-rifai: I like the distinctions james has made here, especially giving the examples, I think having the use cases really helps show off when you might use each type. I like the jeremy's language of "string" better though, since I don't think the term "categorical" variable is very common and I often get asked to explain exactly what that means and why it would be useful.

stephanie-rifai: Though, I think we can drop the collection threads example with only one bound

bradbeam (approved):

PR #205, opened 2020-10-22
Switch CI images to GitHub Container Registry

jgustie:
Making this thing more self contained.


jgustie:
For the life of me, I can't figure out where these images are going. I want to do this if the images are public, but it looks like the three images are created private by default and they aren't showing up as Packages where you change the visibility.

jgustie:
...found them. They were private to Butch; they are public now. This is good to go.

bradbeam (approved):

PR #206, opened 2020-10-22
Expose failure reason and message to the get command

jgustie:


redskyctl/internal/commands/experiments/experiments.go
```diff
@@ -378,6 +386,7 @@ func (m *experimentsMeta) Header(outputFormat string, column string) string {
 	if strings.ToLower(outputFormat) == "csv" {
 		return column
 	}
+	column = regexp.MustCompile("(.)([A-Z])").ReplaceAllString(column, "$1 $2")
```
jgustie, line 389: Don't use Go conventions for camelCase on field names...

bradbeam (approved):

PR #207, opened 2020-10-23
Add the Push Gateway URL to the trial job

jgustie:
What do you think of this? Basically if we see the Prometheus setup task, we add a `PUSHGATEWAY_URL` environment variable to the trial job so inside the trial job all you need to do is `echo "foo 1" | curl --data-binary @- "$PUSHGATEWAY_URL"` to push metrics out (labeled with the trial name).

There is weird circular dependency with setup and trial that I had to switch around so I could properly check to see if the setup task image was the default image or not.

### DIFF
--- a/internal/setup/setup.go
+++ b/internal/setup/setup.go
@@ -18,6 +18,7 @@ package setup
 
 import (
 	"fmt"
+	"strings"
 
 	redskyv1beta1 "github.com/redskyops/redskyops-controller/api/v1beta1"
 	batchv1 "k8s.io/api/batch/v1"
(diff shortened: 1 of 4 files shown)


internal/setup/setup.go
```diff
@@ -142,3 +143,36 @@ func GetConditionStatus(j *batchv1.Job) (corev1.ConditionStatus, string) {
...
+	}
+
+	return env
+}
+
+// AppendPrometheusEnv appends environment variables to help reference the built in Prometheus
+func AppendPrometheusEnv(t *redskyv1beta1.Trial, env []corev1.EnvVar) []corev1.EnvVar {
+	for i := range t.Spec.SetupTasks {
+		if IsPrometheusSetupTask(&t.Spec.SetupTasks[i]) {
+			url := "http://prometheus:9091/metrics/trial/" + t.Name
```
bradbeam: The url here needs to change to be in line with what we deploy -- `redsky-${namespace}-prometheus`

bradbeam (dismissed):
like the layout of this

bradbeam (approved):

PR #208, opened 2020-10-23
Enforce metric bounds

jgustie:
This gives us the ability to specify bounds on metric values. Trials with metrics out of bounds will be marked as failed.


controllers/metric_controller.go
```diff
@@ -191,6 +192,16 @@ func (r *MetricReconciler) collectMetrics(ctx context.Context, t *redskyv1beta1.
 		return controller.RequeueConflict(err)
 	}
 
+	// Wait until all metrics have been collected to fail the trial for an out of bounds metric
+	for i := range t.Spec.Values {
+		v := &t.Spec.Values[i]
+		if err := validation.CheckMetricBounds(metrics[v.Name], v); err != nil {
+			trial.ApplyCondition(&t.Status, redskyv1beta1.TrialFailed, corev1.ConditionTrue, "MetricBound", err.Error(), probeTime)
```
bradbeam, line 199: does failing a trial here prevent us from posting the metrics?  
I'm assuming the metrics would still get sent since the metrics have already been evaluated and exist on the trial

api/v1alpha1/experiment_types.go
```diff
@@ -88,6 +88,10 @@ type Metric struct {
 	Name string `json:"name"`
 	// Indicator that the goal of the experiment is to minimize the value of this metric
 	Minimize bool `json:"minimize,omitempty"`
+	// The inclusive minimum allowed value for the metric
+	Min *resource.Quantity `json:"min,omitempty"`
```
bradbeam, line 92: Do we really want to use `resource.Quantity` for this instead of a float ( since that is what metrics evaluate to)?
Slightly concerned about people putting in something like `100m`

bradbeam (dismissed):

controllers/metric_controller.go
jgustie: It does prevent us from posting the actual metric values to the API (we only send the actual values if the trial is not failed). I don't know if the backend would be OK with us sending failed trials that include values (but maybe that should also be part of non-optimized metrics?). @jdplatt ?

api/v1alpha1/experiment_types.go
jgustie: We can't use floats in CRDs, this was the best compromise. Nothing stops people from specifying `100m` as a bound. I did crank up the scale of the quantity though, by using nano scale instead of milli scale you can specify what appears to be more precise decimal numbers.

bradbeam (approved):

PR #210, opened 2020-10-26
Fix browser error message on login

jgustie:
This changes the error message from whatever `exec` reports to be something more useful.

The reason we don't want to silently ignore the error is that `redskyctl` is running an HTTP server on `localhost` and if they don't have a browser that can hit that localhost server when Auth0 redirects them, the authorization flow will not complete.

When using `redskyctl login --url` we use a device flow instead of an authorization code flow, with the device flow the browser and `redskyctl` do not need to be on the same system (which is more likely the case when opening the browser fails).

### DIFF
--- a/redskyctl/internal/commands/login/login.go
+++ b/redskyctl/internal/commands/login/login.go
@@ -336,14 +336,19 @@ func (o *Options) openBrowser(loc string) error {
 		return err
 	}
 
-	// Do not open the browser for root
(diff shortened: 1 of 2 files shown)


jgustie:
Also fixed `redskyctl results`, including the necessary adjustment to the `--idle-timeout`.

bradbeam (approved):

PR #212, opened 2020-10-29
feat: Add configmap reloader to prometheus deployment 

bradbeam:


config/prometheus/prometheus-server-deployment.yaml
```diff
@@ -74,8 +74,16 @@ spec:
...
+          - --web.enable-lifecycle
         ports:
           - containerPort: 9090
+        resources:
+          requests:
+            cpu: 100m
+            memory: 100M
+          limits:
+            cpu: 200m
+            memory: 1G
```
jgustie, line 86: Is this dependent on #211?

jgustie (approved):

config/prometheus/prometheus-server-deployment.yaml
bradbeam: Should be independent

PR #216, opened 2020-10-30
Suggest trials from the controller

jgustie:
This allows trials to be created in cluster which will be immediately removed after being sent to the server as a suggestion.


jgustie:
This was an idea I had working on experiment generation. During the scanning process, we can actually determine what the current values are: this change would allow us to create a manifest for a "baseline" trial resource during the scan to include with the experiment definition. The controller would see the trial (labelled as a suggestion) and would send the assignments to the server; then it would delete the in-cluster trial (since the server would now have it in the queue).

jgustie:
I'm going with @jdplatt idea to bring back the status-quo field on the parameter definition.

PR #217, opened 2020-11-04
feat: Push release images to ghcr

bradbeam:
Signed-off-by: Brad Beam <brad.beam@carbonrelay.com>

PR #218, opened 2020-11-04
Push release images to ghcr

jgustie:
I'd have done this as a branch off a branch but it would have been a branch off a branch off a fork :trollface: 

It's just compaction of what you did in #217 ...

### DIFF
--- a/.github/workflows/master.yaml
+++ b/.github/workflows/master.yaml
@@ -70,13 +70,13 @@ jobs:
           else
             DOCKER_REGISTRY="ghcr.io/redskyops"
             DOCKER_TAG="sha-$(git rev-parse --short HEAD)"
-            echo "IMAGE_TAG=edge"  >> $GITHUB_ENV
-            echo "${{ secrets.BMASTERS_TOKEN }}" | docker login "ghcr.io" -u "${{ secrets.BMASTERS_ACTOR }}" --password-stdin
+            echo "IMAGE_TAG=edge" >> $GITHUB_ENV
           fi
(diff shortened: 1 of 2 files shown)


.github/workflows/master.yaml
```diff
@@ -90,13 +90,21 @@ jobs:
...
-          docker push "${SETUPTOOLS_IMG%%:*}:${IMAGE_TAG}"
+      - name: Push additional Docker tags
+        run: |
+          docker-push() {
+            docker tag "$1" "$2"
+            docker push "$2"
+          }
+          docker-push "${IMG}" "${IMG%%:*}:${IMAGE_TAG}"
+          docker-push "${REDSKYCTL_IMG}" "${REDSKYCTL_IMG%%:*}:${IMAGE_TAG}"
+          docker-push "${SETUPTOOLS_IMG}" "${SETUPTOOLS_IMG%%:*}:${IMAGE_TAG}"
```
bradbeam, line 100: Should we drop these in favor of the `make docker-push` above, or drop the `Push Docker Images` step?

jgustie: I was trying to break it into steps to avoid documentation; the "Push Docker images" pushes the images used by the product and the "Push additional Docker tags" just pushes the extra tags (e.g. `latest`/`edge`). Though I guess it also pushes to the dev registry so maybe the name wasn't the best...

jgustie: I went back to the one "Push Docker images" and made them the same for master and PRs (master has the extra step of forcing a GitHub Container Registry push).

bradbeam (approved):

PR #219, opened 2020-11-05
Add a special case to RBAC generation for Elastic

jgustie:
The reason for the resource mismatch we saw today was our reliance on the "unsafe guess" for converting "Kind" to "Plural Resource". The unsafe guess just adds an "s" (or replaces "y" with "ies"). This workaround registers the type so we map it properly without falling back to unsafe guess. As we add more of these, we might want to externalize the extra mappings somewhere else (e.g. in code or in the configuration). Another option to consider is making it user configurable from the CLI.

### DIFF
--- a/redskyctl/internal/commands/generate/rbac.go
+++ b/redskyctl/internal/commands/generate/rbac.go
@@ -32,6 +32,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/rand"
 	"k8s.io/client-go/kubernetes/scheme"
 )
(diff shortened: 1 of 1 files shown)


redskyctl/internal/commands/generate/rbac.go
```diff
@@ -90,9 +91,21 @@ func NewRBACCommand(o *RBACOptions) *cobra.Command {
 func (o *RBACOptions) Complete() {
 	// Create a REST mapper to convert from GroupVersionKind (used on patch targets) to GroupVersionResource (used in policy rules)
 	rm := meta.NewDefaultRESTMapper(scheme.Scheme.PreferredVersionAllGroups())
+
+	// Add all types known to the Kubernetes Client Go scheme (we can use unsafe guess safely for these types)
 	for gvk := range scheme.Scheme.AllKnownTypes() {
 		rm.Add(gvk, meta.RESTScopeRoot)
 	}
+
+	// If we hit a CRD we will use "unsafe guess" (basically add an "s"); here are some well-known exceptions
```
bradbeam: Should we shell out to kubectl and fill in the info?
```
kubectl get crd experiments.redskyops.dev -o go-template='{{ .spec.names.plural }}'
kubectl get crd experiments.redskyops.dev -o go-template='{{ .spec.names.singular }}'
```

jgustie: I like it.

bradbeam (approved):

PR #220, opened 2020-11-05
Add parameter baseline configuration

jgustie:
This PR adds the ability to specify a baseline configuration in the experiment. It does not impact the current approach of labeling a baseline. If even a single parameter does not have a baseline, we will not send a suggestion.


internal/server/server.go
```diff
@@ -132,8 +148,13 @@ func FromCluster(in *redskyv1beta1.Experiment) (redskyapi.ExperimentName, *redsk
 		})
 	}
 
+	// We need all of the assignments to include the baseline
+	if len(baseline.Assignments) == 0 || len(baseline.Assignments) != len(out.Parameters) {
```
bradbeam, line 152: Should we surface an error here if `len(baseline.Assignments) != len(out.Parameters)`?

bradbeam (dismissed):

internal/server/server.go
jgustie: I was just silently ignoring it and treating it as a nice to have. Maybe I can add something to `check experiment`...

bradbeam (approved):

PR #224, opened 2020-11-12
Include build metadata in UA string of pre-releases

jgustie:
The answer to another of my TODO question is "yes".

### DIFF
--- a/internal/version/useragent.go
+++ b/internal/version/useragent.go
@@ -63,13 +63,34 @@ func userAgentString(product, comment string) string {
 	if product == "" {
 		return ""
 	}
-	// TODO Validate "product"
 
+	// TODO Validate "product"
+	ua := strings.Builder{}
+	ua.WriteString(product)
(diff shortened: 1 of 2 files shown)


internal/version/useragent.go
```diff
@@ -63,13 +63,34 @@ func userAgentString(product, comment string) string {
...
+	// Only include build metadata for pre-release versions
+	if strings.Contains(Version, "-") && BuildMetadata != "" {
+		comments = append(comments, BuildMetadata)
+	}
 
+	// Clean white space and surrounding comment indicators
+	comment = strings.TrimSpace(comment)
+	comment = strings.TrimLeft(comment, "(")
+	comment = strings.TrimRight(comment, ")")
 	comment = strings.TrimSpace(comment)
```
bradbeam, line 85: should these be under the if statement below?

bradbeam (dismissed):

bradbeam (approved):

PR #227, opened 2020-11-13
fix: Properly set GroupKindVersion on experiment resource

bradbeam:


jgustie:
You should never need to set type meta on a registered Go type, what is this fixing?

bradbeam:
we'll just classify this under bugs that come up at 2am.

PR #229, opened 2020-11-16
Adjust the query time to account for the last scrape interval

jgustie:
We are seeing that especially with the push gateway, Prometheus may not have data present at the completion time of the trial, instead the data is coming in on the subsequent scrape. This fix pushes the query time back to account for the scrape intervals.


jgustie:
The one problem I see with this is that we are not necessarily getting _just_ the scrape directly after the completion of the trial depending on the offsets of scrape times on the targets (e.g. if a target can be scraped twice prior to the successful scrape of another target, we will pick up both the scrapes of the fast target waiting for the slow target). It may be worthwhile limiting `queryTime` to `completionTime + d` where `d` is something like 2x the scrape interval (which we do not currently know in this code).

bradbeam (approved):

PR #232, opened 2020-11-17
fix: Update login urls

bradbeam:


jgustie:
I don't want to merge any URL changes in the controller until we know the backend is correctly handling the redirects otherwise we are just masking issues with current versions of the product.

jgustie (approved):

PR #234, opened 2020-11-17
Deprecate results

jgustie:
Remove the UI from the CLI.


redskyctl/internal/commands/results/results.go
```diff
@@ -17,157 +17,65 @@ limitations under the License.
...
-}
-
-func (o *Options) openBrowser(loc string) error {
+func (o *Options) results() error {
 	u, err := user.Current()
 	if err != nil {
 		return err
 	}
 
+	loc := "https://app.stormforge.io/experiments"
```
bradbeam, line 67: Should we/can we pull this in from the config? More of a dev issue

bradbeam (approved):

redskyctl/internal/commands/results/results.go
jgustie: I added configuration to `optimize-go`; I will pick it up here and in #232 once it is merged and released.

PR #238, opened 2020-11-19
fix: Patch trial job instead of delete it

bradbeam:
We can suspend/interrupt a pending trial job by setting the parallelism to 0.
This should allow us to preserve trial run history and handle cases where
the trial job fails to schedule because of resource constraints.

Signed-off-by: Brad Beam <brad.beam@b-rad.info>

### DIFF
--- a/controllers/trial_job_controller.go
+++ b/controllers/trial_job_controller.go
@@ -18,6 +18,7 @@ package controllers
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	"github.com/go-logr/logr"
@@ -29,6 +30,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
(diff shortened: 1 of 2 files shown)


controllers/trial_job_controller.go
```diff
@@ -66,23 +68,25 @@ func (r *TrialJobReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+		return ctrl.Result{}, nil
+	}
+
+	// Insert a "sleep" between "ready" and the trial job
+	if ids := time.Duration(t.Spec.InitialDelaySeconds) * time.Second; ids > 0 {
+		for _, c := range t.Status.Conditions {
+			if c.Type == redskyv1beta1.TrialReady {
+				startTime := c.LastTransitionTime.Add(ids)
+				if startTime.After(now.Time) {
+					return ctrl.Result{RequeueAfter: startTime.Sub(now.Time)}, nil
```
bradbeam, line 81: This was more of a cosmetic change and can be reverted if needed. Wanted to unwrap the code block by returning early.

controllers/trial_job_controller.go
```diff
@@ -179,17 +183,24 @@ func (r *TrialJobReconciler) applyJobStatus(ctx context.Context, t *redskyv1beta
...
+
 				// TODO We should consolidate this with `internal/ready/podFailed`
 				for _, c := range s.Conditions {
 					if c.Type == corev1.PodScheduled && c.Status == corev1.ConditionFalse && c.Reason == corev1.PodReasonUnschedulable {
-						trial.ApplyCondition(&t.Status, redskyv1beta1.TrialFailed, corev1.ConditionTrue, c.Reason, c.Message, time)
+						trial.ApplyCondition(&t.Status, redskyv1beta1.TrialFailed, corev1.ConditionTrue, c.Reason, fmt.Sprintf("trial pod: %s", c.Message), time)
+
+						// Patch the job and set parallelism to 0 to suspend the job and terminate any active pods
+						if err := r.Patch(ctx, job, client.RawPatch(types.StrategicMergePatchType, []byte(`{ "spec": { "parallelism": 0  } }`))); err != nil {
+							r.Log.Error(err, "unable suspend trial job", "job", job, "pod status", podList.Items[i].Status)
```
jgustie: Nit: the messages should be title case (unlike errors which should be lower case; unlike errors, the messages are presented directly). Also, I get nervous about putting full objects into the log; at this point I think it would be sufficient to have `WithValues("trial", fmt.Sprintf("%s/%s", t.Namespace, t.Name))` (matches what we do in the metric controller) and `WithValues("job", job.Name)`.

jgustie (approved):

PR #244, opened 2020-11-25
Metric URLs, part deux

jgustie:
This is a slightly less involved refactoring then the first pass (though there is still a bit of work here). Instead of trying to preserve round trips of new data to the old format, this refactoring just ignores the problem. It also ignores named ports. There is a reasonable probability that `v1alpha1` metrics will _require_ refactoring to become usable again (though the built-in Prometheus metrics should be all set; which, if we are being honest probably covers 90% because if you were forced to use Prometheus via a service you probably already want to switch to the URL).

The big changes:
1. The "local" and "pods" metric types are consolidated into one "kubernetes" type. The "kubernetes" type allows you to query the Kube API for any kind by name or label. If you do not specify a kind to query, the default behavior is to just return the trial object itself. This is maybe useful if you have metrics published in CRDs (or want durations of specific pods), but as of right now controller/experiment RBAC generation _does not_ account for this so you need to make sure the controller has the right permissions.
2. All other metric types are now "URL" metrics: they are based on a URL and ignore the Kube API kind/selector. There are otherwise no changes to these metric types. As a special case to keep as many existing metrics functional as possible, conversion will create a URL with the host name `redskyops.dev` which will trigger a service look up with a subset of the old `toURL` logic. This is intended as a temporary bridge to get people over to the URL _before_ we rename the CRD. Once we rename the CRD (likely in the next release) this code will be removed.
3. Metric collection is using `NaN` to allow reporting of 0 valued error distribution (currently explicit 0 is not allowed because it is the "nothing to report" case). This does not impact what we send to the API, it is a change to the internal sentinel value only.
4. There is a new `indexResource` function for templates which behaves like the standard `index` function except is actually works on a `ResourceList` and returns a pointer to a quantity instead of a value (so the Go template can choose a scale). Technically this was only necessary to get the test working, but it seems like a good idea to have.

5. Late addition: the combination of `ObjectReference` and `LabelSelector` fields has been replaced by a `ResourceTarget` type to help clarify the relation between the selector and the GVK. This will be picked in more places in the API in a later PR.


jgustie:
Still a draft as I haven't done much testing on this yet.

api/v1alpha1/experiment_conversion.go
```diff
@@ -91,7 +95,126 @@ func Convert_v1beta1_Parameter_To_v1alpha1_Parameter(in *v1beta1.Parameter, out
...
+
+	default:
+		// Ignore types that are not URL based
+		return ""
+	}
+
+	// Use a special placeholder for the host
+	u := url.URL{
+		Scheme: m.Scheme,
+		Host:   "redskyops.dev",
```
bradbeam: we should make this a constant somewhere

controllers/metric_controller.go
```diff
@@ -138,11 +142,13 @@ func (r *MetricReconciler) evaluateMetrics(ctx context.Context, t *redskyv1beta1
...
 
 func (r *MetricReconciler) collectMetrics(ctx context.Context, t *redskyv1beta1.Trial, probeTime *metav1.Time) (*ctrl.Result, error) {
-	// Index the metric definitions
+	// Fetch the experiment
 	exp := &redskyv1beta1.Experiment{}
 	if err := r.Get(ctx, t.ExperimentNamespacedName(), exp); err != nil {
 		return &ctrl.Result{}, err
 	}
+
+	// Index the metric definitions
```
bradbeam: should the metrics collection be ( close to ) atomic or should we allow partial metrics to be collected as we go?

ex, if I have 5 metrics configured and the 5th metric is giving me an issue, does it make sense to recollect the previous 4 metrics each time or should we stash each metric in the status as we get the results?

controllers/metric_controller.go
```diff
@@ -138,11 +142,13 @@ func (r *MetricReconciler) evaluateMetrics(ctx context.Context, t *redskyv1beta1
...
 	metrics := make(map[string]*redskyv1beta1.Metric, len(exp.Spec.Metrics))
 	for i := range exp.Spec.Metrics {
 		metrics[exp.Spec.Metrics[i].Name] = &exp.Spec.Metrics[i]
```
bradbeam: should we do `metricWithDefaults` here to generate the DeepCopy early?

controllers/metric_controller.go
```diff
@@ -156,40 +162,34 @@ func (r *MetricReconciler) collectMetrics(ctx context.Context, t *redskyv1beta1.
...
-			if stddev != 0 {
-				v.Error = strconv.FormatFloat(stddev, 'f', -1, 64)
-			}
+		// Use a defaulted DEEP COPY of the metric so we can safely make changes
+		m := metricWithDefaults(t, metrics[v.Name])
+
+		// This is strictly for converted v1alpha1 experiments
+		if err := r.resolveLegacyURL(ctx, t, m); err != nil {
+			return r.collectionAttempt(ctx, log, t, v, probeTime, err)
 		}
```
bradbeam: probably just me ( so maybe uber-nit?) , but I'd put this with `metricWithDefaults`

controllers/metric_controller.go
```diff
@@ -208,49 +208,131 @@ func (r *MetricReconciler) collectMetrics(ctx context.Context, t *redskyv1beta1.
...
+	trial.ApplyCondition(&t.Status, redskyv1beta1.TrialObserved, corev1.ConditionFalse, "", "", probeTime)
 
-		if m.URL != "" {
-			return target, nil
-		}
+	// Decrement the attempts remaining, if there are no attempts left, fail the trial
+	if v.AttemptsRemaining > 0 {
+		v.AttemptsRemaining--
+		if v.AttemptsRemaining == 0 {
+			trial.ApplyCondition(&t.Status, redskyv1beta1.TrialFailed, corev1.ConditionTrue, "MetricFailed", err.Error(), probeTime)
```
bradbeam: nit, I think this would still be the same but would save a level
```
if v.AttemptsRemaining > 0 {
  v.AttemptsRemaining--
}
if v.AttemptsRemaining == 0 {
  metric failed.. 
}
```

internal/metric/metric.go
```diff
@@ -17,133 +17,37 @@ limitations under the License.
...
+		return value, math.NaN(), err
 	case redskyv1beta1.MetricPrometheus:
-		return capturePrometheusMetric(log, metric, target, trial.Status.CompletionTime.Time)
+		return capturePrometheusMetric(ctx, log, metric, trial.Status.CompletionTime.Time)
 	case redskyv1beta1.MetricDatadog:
-		return captureDatadogMetric(metric.Scheme, metric.Query, trial.Status.StartTime.Time, trial.Status.CompletionTime.Time)
+		return captureDatadogMetric(metric, trial.Status.StartTime.Time, trial.Status.CompletionTime.Time)
 	case redskyv1beta1.MetricJSONPath:
-		return captureJSONPathMetric(metric, target)
+		return captureJSONPathMetric(metric)
```
bradbeam, line 49: this seems to surface the pattern of a metric target/collector being `func(ctx, log, metric, start time, end time)`; might be a good candidate for an interface.

internal/metric/prometheus.go
```diff
@@ -28,103 +28,103 @@ import (
...
+	Message string
+	// The URL that was used to capture the metric
+	Address string
+	// The metric query that failed
+	Query string
+	// The completion time at which the query was executed
+	CompletionTime time.Time
+	// The minimum amount of time until the metric is expected to be available
+	RetryAfter time.Duration
+}
```
bradbeam: I'm a bit torn on this being specific to prometheus

bradbeam (commented):
some good cleanup in this.

controllers/metric_controller.go
jgustie: Attempts remaining will be zero coming in on a successful collection attempt, it's only a failure if you decrease to zero.

jgustie: ...you could flip it and check for 1:
```
if v.AttemptsRemaining == 1 {
  metric failed.. 
}
if v.AttemptsRemaining > 0 {
  v.AttemptsRemaining--
}
```

controllers/metric_controller.go
jgustie: How about if I just make this do the deep copy while indexing and fill in the combined defaults/legacy URL in the loop below?

(the point of not doing defaults up here was that there is no point in computing defaults on completed metrics)

controllers/metric_controller.go
jgustie: The reason it's not is that resolution of the legacy URL requires state from the reconciler (the client), while the current defaults implementation is a global function. I'll adjust things by making defaults and legacy URL resolution more coupled.

jgustie:
There is a bug in some of the template logic for `type: pods` that I am working on.

controllers/metric_controller.go
```diff
@@ -208,49 +207,145 @@ func (r *MetricReconciler) collectMetrics(ctx context.Context, t *redskyv1beta1.
...
-		}
+	// Update the probe time and ensure that trial observed is still explicitly false (i.e. we have started observation but it is not complete)
+	trial.ApplyCondition(&t.Status, redskyv1beta1.TrialObserved, corev1.ConditionFalse, "", "", probeTime)
 
-		return target, nil
-	case redskyv1beta1.MetricPrometheus, redskyv1beta1.MetricJSONPath:
-		// Both Prometheus and JSONPath target a service
-		target := &corev1.ServiceList{}
+	// If there was only one attempt remaining, mark the trial as failed
+	if v.AttemptsRemaining == 1 {
```
bradbeam: I dont think we'd run into this situation, but should we do  `v.AttemptsRemaining <= 1`

bradbeam: Or should we rephrase this to always decrement

```
v.AttemptsRemaining--

if v.AttemptsRemaining <= 0 {
		trial.ApplyCondition(&t.Status, redskyv1beta1.TrialFailed, corev1.ConditionTrue, "MetricFailed", err.Error(), probeTime)
		v.AttemptsRemaining = 0

		// Metric errors contain additional information which should be logged for debugging
		if merr, ok := err.(*metric.CaptureError); ok {
			log.Error(merr, "Metric collection failed", "address", merr.Address, "query", merr.Query, "completionTime", merr.CompletionTime)
		}
}
```

controllers/metric_controller.go
```diff
@@ -208,49 +207,145 @@ func (r *MetricReconciler) collectMetrics(ctx context.Context, t *redskyv1beta1.
...
+func (r *MetricReconciler) target(ctx context.Context, t *redskyv1beta1.Trial, m *redskyv1beta1.Metric) (runtime.Object, error) {
+	if m.Type != redskyv1beta1.MetricKubernetes && m.Type != "" {
+		return nil, nil
+	}
+
+	// Get the target reference, default to the trial itself if there is no reference
+	targetRef := m.Target
+	if m.Target == nil || m.Target.Kind == "" {
+		return t, nil
+	}
```
bradbeam: more of a nit, but do we want to do the assignment after the test or change the test to use the assignment ( targetRef ).

controllers/metric_controller.go
```diff
@@ -208,49 +207,145 @@ func (r *MetricReconciler) collectMetrics(ctx context.Context, t *redskyv1beta1.
...
+// applyMetricDefaults fills in default values for the supplied metric.
+func (r *MetricReconciler) applyMetricDefaults(ctx context.Context, t *redskyv1beta1.Trial, m *redskyv1beta1.Metric) error {
+	// Give Prometheus metrics a default URL
+	if m.Type == redskyv1beta1.MetricPrometheus && m.URL == "" {
+		m.URL = fmt.Sprintf("http://redsky-%s-prometheus:9090/", t.Namespace)
+	}
+
+	// Default to the trial namespace
+	if m.Target != nil && m.Target.Namespace == "" {
+		// TODO How do we exclude non-namespaced kinds from this default?
```
bradbeam: should we drop this TODO

testing via kubectl, `kubectl get clusterole` vs `kubectl get clusterrole -n kube-system` both seem to work

controllers/metric_controller.go
```diff
@@ -208,49 +207,145 @@ func (r *MetricReconciler) collectMetrics(ctx context.Context, t *redskyv1beta1.
...
+// looking up a Kubernetes Service object. This roughly corresponds to the original behavior of the controller
+// where URL based metrics were defined using Service selectors instead of actual URLs.
+func (r *MetricReconciler) resolveLegacyURL(ctx context.Context, t *redskyv1beta1.Trial, m *redskyv1beta1.Metric) error {
+	if m.Type != redskyv1beta1.MetricPrometheus && m.Type != redskyv1beta1.MetricJSONPath {
+		return nil
+	}
+
+	// Look for the special placeholder hostname that indicates we should look up a service
+	u, err := url.Parse(m.URL)
+	if err != nil || u.Hostname() != redskyv1alpha1.LegacyHostnamePlaceholder {
```
bradbeam: nit, should we split this up?

```
if err != nil {
return err
}

if u.Hostname() != redskyv1alpha1.LegacyHostnamePlaceholder {
return nil
}
```
I know what's here currently works, but this makes it very clear that LegacyHostname is not actually an error condition

bradbeam (dismissed):

controllers/metric_controller.go
jgustie: We actually don't even need `targetRef` anymore.

bradbeam (dismissed):

bradbeam (approved):

PR #247, opened 2020-12-03
Do not fail baseline trials for having out of bounds metric values

jgustie:
Before enforcing metric bounds we check to see if the trial represents a baseline: either because it is explicitly labeled as such or because it has assignments matching the baseline definition of the experiment.

### DIFF
--- a/internal/trial/trial.go
+++ b/internal/trial/trial.go
@@ -23,6 +23,7 @@ import (
 	redskyv1beta1 "github.com/thestormforge/optimize-controller/api/v1beta1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
 // IsFinished checks to see if the specified trial is finished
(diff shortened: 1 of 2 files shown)


internal/trial/trial.go
```diff
@@ -90,6 +91,38 @@ func IsTrialJobReference(t *redskyv1beta1.Trial, ref *corev1.ObjectReference) bo
 	return true
 }
 
+// IsBaseline checks to see if the supplied trial is a baseline for an experiment.
+func IsBaseline(t *redskyv1beta1.Trial, exp *redskyv1beta1.Experiment) bool {
```
bradbeam, line 95: Dont we do something similar to this elsewhere?

bradbeam (approved):

internal/trial/trial.go
jgustie: The closest thing is [`server.FromCluster`](https://github.com/thestormforge/optimize-controller/blob/master/internal/server/server.go#L45) however that generates an API object so instead of `IntOrStr` (which we can do a Go `==` check on) it has `NumberOrString`.

jgustie:
I'm told there is a JIRA ticket for tracking that this is in fact safe to do.

PR #250, opened 2020-12-04
Rename "labelSelector" and "metricSelector" to "selector"

jgustie:
This PR changes the way utilization queries are rendered to allow for more expressive PromQL metric selectors to be passed through. It also changes some of the variable names to align with the names used by Prometheus.


internal/template/utilization_functions.go
```diff
@@ -72,69 +73,108 @@ scalar(
...
+	// TODO We should remove this
+	// In the case of our built-in Prometheus, we can just alter the scape configs to not include the prefix
+	// In the case of a managed Prometheus we won't have control over this prefix and it may interfere with metric selection
+	if !strings.HasPrefix(key, "label_") {
+		key = "label_" + key
+	}
+
+	// TODO We should remove this
+	// Instead of sanitizing, we should just fail if the key contains invalid values
+	key = invalidLabelCharRE.ReplaceAllString(key, "_")
```
bradbeam: Would we catch this instead in `redskyctl check experiment`?

Given how we're separating things in the application I think it is an okay change, but it causes some end user friction in that you need to mentally be aware that kube labels aren't exactly equal to prometheus labels.

internal/template/utilization_functions.go
```diff
@@ -72,69 +73,108 @@ scalar(
...
-	// Always include the trial namespace
-	labels := []string{fmt.Sprintf("namespace=\"%s\"", data.Trial.Namespace)}
+	// Allow the extra label matchers to be variadic or comma-delimited
+	extraLabelMatchers = strings.Split(strings.Join(extraLabelMatchers, ","), ",")
 
-	for _, label := range strings.Split(strings.Join(labelArgs, ","), ",") {
-		if label == "" {
+	// Always include the trial namespace first
+	labelMatchers := make([]string, 0, 1+len(extraLabelMatchers))
+	labelMatchers = append(labelMatchers, fmt.Sprintf("namespace=\"%s\"", metricData.Trial.Namespace))
```
bradbeam: should we do `namespace=%q` instead?

internal/template/utilization_functions.go
```diff
@@ -72,69 +73,108 @@ scalar(
...
+	if !strings.HasPrefix(value, `"`) {
+		value = strconv.Quote(value)
+	}
+
+	// TODO We should remove this
+	// In the case of our built-in Prometheus, we can just alter the scape configs to not include the prefix
+	// In the case of a managed Prometheus we won't have control over this prefix and it may interfere with metric selection
+	if !strings.HasPrefix(key, "label_") {
+		key = "label_" + key
+	}
```
bradbeam: this specifically comes in for how kube state metrics handles the translation of the pod labels into prometheus labels. 
given where this comes into play and the dependency on kube state metrics, im not sure it's worth having the `TODO` around removing this.

internal/template/utilization_functions.go
```diff
@@ -72,69 +73,108 @@ scalar(
...
+		key, op, value, err := splitLabelMatcher(labelMatcher)
+		if err != nil {
+			return "", err
 		}
 
-		// Sanitize label name for prometheus usage
-		kvpair[0] = invalidLabelCharRE.ReplaceAllString(kvpair[0], "_")
-
-		labels = append(labels, fmt.Sprintf("label_%s=\"%s\"", kvpair[0], kvpair[1]))
+		labelMatchers = append(labelMatchers, key+op+value)
```
bradbeam: If/when we take out the additional massaging of labels in `splitLabelMatcher`, will `splitLabelMatcher` turn more into a  validation step ( as in return bool ) and then we append `labelMatcher` directly to `labelMatchers` instead of reassembling it?

bradbeam (dismissed):

internal/template/utilization_functions.go
jgustie: We can always relabel metrics however we want in the built-in Prometheus to strip the `label_` prefix ; and for a managed Prometheus we can't possibly know what relabel configs they have applied, so even if that's the convention there is no guarantee that is what they are using.

internal/template/utilization_functions.go
jgustie: I would prefer to make it a straight pass through and get rid of `splitLabelMatcher` altogether (just join the variadic args coming in from the template function using a comma; and even that is really just a concession to generalize making the metric selector optional). We can do some additional validation in the linter as you suggested to check the syntax.

internal/template/utilization_functions.go
```diff
@@ -75,67 +74,95 @@ scalar(
...
+	}
 
-		kvpair := strings.Split(label, "=")
-		if len(kvpair) != 2 {
-			return "", fmt.Errorf("invalid label for utilization query, expected key=value, got: %s", label)
+	// Always include the trial namespace first
+	requirements, _ := sel.Requirements()
+	labelMatchers := make([]string, 0, 1+len(requirements))
+	labelMatchers = append(labelMatchers, fmt.Sprintf("namespace=%q", metricData.Trial.Namespace))
+	for _, req := range requirements {
```
bradbeam, line 119: Should we filter out namespace? Kind of a weird/special case since we add it in, but if someone were to specify it in a query we'd try to wrap it with label_.

bradbeam (commented):
Should be fine after chat about namespace handling

internal/template/utilization_functions.go
jgustie: I think `label_namespace` should be the behavior here, we are only trying to let the user query Prometheus metrics based on Kubernetes labels. I think we should stick to generating the `label_` only queries and if people need to query on other Prometheus metric labels, they can do that using their own PromQL.

jgustie:
Grrrr. I forgot the parameter list was branch off of this when I merged it.

bradbeam (approved):

PR #251, opened 2020-12-04
Add default TTL for server based trials

jgustie:
This add a default 48 hour TTL to trials that were suggested by the server (we don't default the TTL for manually created trials because the only record of their existence is in the cluster, we also don't default the TTL for trials whose approximate runtime is greater then 48 hours). 

After this change you can still keep trials in your cluster indefinitely, even for server suggested trials, by explicitly setting the `ttlSecondsAfterFinished` on the trial template to `-1` (a TTL of 0 indicates the trial should be immediately deleted).

### DIFF
--- a/controllers/server_controller.go
+++ b/controllers/server_controller.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"strings"
+	"time"
 
 	"github.com/go-logr/logr"
 	redskyv1beta1 "github.com/thestormforge/optimize-controller/api/v1beta1"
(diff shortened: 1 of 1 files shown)


jgustie:
I still don't know how I feel about this. I'm not a huge fan of default automatic clean up; it can be unexpected in very unpleasant ways.

bradbeam (approved):

bradbeam:
> I still don't know how I feel about this. I'm not a huge fan of default automatic clean up; it can be unexpected in very unpleasant ways.

I think what's pushing me forward with this is that we can get the same outcome with redskyctl get trials -- `redskyctl get trials postgres-example -o csv`. I feel like that's often neglected today because we rely on `kubectl get trial` instead.

PR #254, opened 2020-12-10
fix: Add duration to kube_pod_labels in utilization queries

bradbeam:
This should expand the search space for kube_state_metrics queries so we can
get back a correct set of labels/pods when we make our query. By increasing
the space we are more tolerant of query time offsets.


bradbeam:
Not sure If I've got the branching set up right for that.

jgustie (dismissed):

jgustie (approved):

PR #258, opened 2020-12-17
Adjust Prometheus query time only if initial result is NaN

jgustie:
This PR makes it so the query time for Prometheus metrics is only adjusted if initial result returns NaN. The idea being that under normal circumstances the completion time _is_ the correct time to query from. When using the Pushgateway, a metric value will be NaN until the scrape after the completion: so this change makes a second query at time of the last scrape following completion in hopes of seeing the values from the Pushgateway.

An alternative solution would have been to introduce a sleep for the length of the scrape interval to all trial jobs pushing metrics (this would have been unreasonable to enforce).

This also takes some of the changes from #244 (in a way that will conflict, I'll fix that after merging) to clean up Prometheus metric collection.

### DIFF
--- a/internal/metric/prometheus.go
+++ b/internal/metric/prometheus.go
@@ -52,6 +52,9 @@ func capturePrometheusMetric(log logr.Logger, m *redskyv1beta1.Metric, target ru
 }
 
 func captureOnePrometheusMetric(log logr.Logger, address, query, errorQuery string, completionTime time.Time) (float64, float64, error) {
+	ctx := context.TODO()
+	var value, valueError float64
+
 	// Get the Prometheus client based on the metric URL
 	// TODO Cache these by URL
 	c, err := prom.NewClient(prom.Config{Address: address})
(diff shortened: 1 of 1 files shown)


internal/metric/prometheus.go
```diff
@@ -61,70 +64,85 @@ func captureOnePrometheusMetric(log logr.Logger, address, query, errorQuery stri
...
-			queryTime = target.LastScrape
-		}
-	}
-
-	// If we needed to adjust the query time, log it so we have a record of the actual time being used
-	if !queryTime.Equal(completionTime) {
-		log.WithValues("queryTime", queryTime).Info("Adjusted completion time for Prometheus query")
-	}
-
 	// Execute query
```
bradbeam, line 72: should we loop here?
```
for _, query := range []string{ query, errorquery } {
  if query == "" { continue }
  <query scalar>
  <adjust>
}
```

bradbeam (approved):

internal/metric/prometheus.go
jgustie: Yeah, that's a fair point, otherwise when we actually start querying for error rates the timing could be off. The completion time probably needs to be adjusted during the loop as well so the error range matches the adjusted trial range if we decided that the final scrape was required. I'm going to make an issue to track this for now.

PR #261, opened 2021-01-06
feat: Migrate jwt libraries

bradbeam:
This consolidates the jwt libraries we use and switches us to a more maintained version.


redskyctl/internal/commands/check/config.go
```diff
@@ -93,12 +93,17 @@ func tenantID(r config.Reader) (string, error) {
 	if az.Credential.TokenCredential == nil {
 		return "", nil
 	}
-	mc := jwt.MapClaims{}
-	if _, _, err := new(jwt.Parser).ParseUnverified(az.Credential.TokenCredential.AccessToken, mc); err != nil {
+
+	token, err := jwt.ParseString(
+		az.Credential.TokenCredential.AccessToken,
+		jwt.WithValidate(true),
```
jgustie: Do not validate the token here, often times the token expires in the configuration file and won't be updated until it's needed (by way of the refresh token). This would cause check config to fail unnecessarily. We are not using the information in the token, we are just presenting it for informational purposes so skipping validation is a reasonable compromise.

redskyctl/internal/commands/check/config.go
```diff
@@ -93,12 +93,16 @@ func tenantID(r config.Reader) (string, error) {
 	if az.Credential.TokenCredential == nil {
 		return "", nil
 	}
-	mc := jwt.MapClaims{}
-	if _, _, err := new(jwt.Parser).ParseUnverified(az.Credential.TokenCredential.AccessToken, mc); err != nil {
+
+	token, err := jwt.ParseString(
+		az.Credential.TokenCredential.AccessToken,
+	)
```
jgustie, line 99: nit: this can probably be one line now

jgustie (approved):

PR #268, opened 2021-01-22
Error rate objective

jgustie:
This PR adds a new objective to the application named "error rate". It can be used with both StormForger and Locust.

To make your experiment optimize for the per-request error rate add an objective named either `error-rate`, `error-ratio`, or `errors`. Alternatively, you can name your objective whatever you like and add `errorRate: requests` to the objective.


txpdev (approved):
Great!

PR #271, opened 2021-01-29
Parameter defaults from baselines

jgustie:
The current experiment generation logic uses a fixed parameter range. It is possible that this will produce an invalid experiment if the baseline value falls outside the the fixed range.

This change adjusts the fixed range when a baseline value is present, see the test for examples.

### DIFF
--- a/redskyctl/internal/commands/generate/experiment/containerresources.go
+++ b/redskyctl/internal/commands/generate/experiment/containerresources.go
@@ -233,8 +233,8 @@ func toIntWithRange(name corev1.ResourceName, q resource.Quantity) (value *intst
 		max = int32(math.Pow(2, math.Ceil(math.Log2(float64(scaled.IntVal*2)))))
 	case corev1.ResourceCPU:
 		scaled = intstr.FromInt(int(q.ScaledValue(resource.Milli)))
-		min = int32((float64(scaled.IntVal) / 200) * 100)
-		max = int32((float64(scaled.IntVal) / 50) * 100)
+		min = int32(math.Floor(float64(scaled.IntVal)/20)) * 10
+		max = int32(math.Ceil(float64(scaled.IntVal)/10)) * 20
 	}
 
 	return &scaled, min, max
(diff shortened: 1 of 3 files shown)


redskyctl/internal/commands/generate/experiment/containerresources.go
```diff
@@ -221,3 +221,21 @@ func materializeResourceList(node *yaml.MapNode) corev1.ResourceList {
...
 }
+
+// toIntWithRange returns the quantity as an int value with a default range.
+func toIntWithRange(name corev1.ResourceName, q resource.Quantity) (value *intstr.IntOrString, min int32, max int32) {
+	var scaled intstr.IntOrString
+
+	switch name {
+	case corev1.ResourceMemory:
+		scaled = intstr.FromInt(int(q.ScaledValue(resource.Mega)))
+		min = int32(math.Pow(2, math.Floor(math.Log2(float64(scaled.IntVal/2)))))
```
bradbeam: This feels clever. 
Am I understanding it right?
min = 2^(val/2 taking care to end on a bit boundary)
max = 2^(val*2 taking care to end on a bit boundary)

bradbeam (dismissed):

redskyctl/internal/commands/generate/experiment/containerresources.go
jgustie: Yeah, just sticking with powers of 2 for memory values but not CPU. Which reminds me: I did not account for replicas.

txpdev: LGTM, my main issue here is that we don't enforce limits in the genereated min max. for example if i set a baseline of 1050 i get max= pow(2, ceil(log2(1050*2)))=4096. Do i really want to have 4GB of memory requested?

txpdev (dismissed):
LGTM, quick comment about enforcing a limit for the generated max

bradbeam (dismissed):

bradbeam (approved):

PR #274, opened 2021-02-03
ConfigMap patch order

jgustie:
This is kind of an unexpected add, environment variables ended up being a tougher add then expected so this bridges the gap.

The ConfigMap scanner works by adding a list of config map key specifications to the application's parameters section. For example:
```
parameters:
  configMaps:
  - name: my-config-map
    key: some-key
    numericValue:
      min: 0
      max: 10
      prefix: x
      suffix: x
    stringValue:
    - x
    - y
    - z
```

Note that `numericValue` and `stringValue` are mutually exclusive, they are both shown here to illustrate the full schema. The prefix/suffix values allow for a numeric value to have additional information in the config map. The above example would identify and patch the following ConfigMap with a baseline of `5` (ignoring the fact that `stringValue` was set):
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: my-config-map
data:
  some-key: x5x
```

The resulting experiment would then include the following patches and parameters:

```
apiVersion: redskyops.dev/v1beta1
kind: Experiment
spec:
  parameters:
  - baseline: 5
    max: 10
    name: some-key
  patches:
  - patch: |
      data:
        some-key: "x{{ .Values.some-key }}x"
    targetRef:
      apiVersion: v1
      kind: ConfigMap
      name: my-config-map
```

Additional notes:
* The range _must_ be explicitly specified, we do not make attempts to guess the range of numeric parameters
* The value in the scanned resource must match the description in the Application. In the above example, had `some-key: b5b` (wrong prefix/suffix) experiment generation would fail.


redskyctl/internal/commands/generate/experiment/configmaps.go
```diff
@@ -0,0 +1,201 @@
...
+	// Data key in the ConfigMap to update.
+	Key string `json:"key,omitempty"`
+	// Create the config map entry even if the original object does not contain it.
+	CreateIfNotPresent bool `json:"create,omitempty"`
+
+	// Specification of a numeric value to optimize. Mutually exclusive with the string specification.
+	NumericValue *redskyappsv1alpha1.NumericParameter `json:"numericValue,omitempty"`
+	// Specification of possible string values. Mutually exclusive with the numeric specification.
+	StringValue redskyappsv1alpha1.StringParameter `json:"stringValue,omitempty"`
+}
```
bradbeam: Should we embed apps.ConfigMap here and add on?
```
type ConfigMapKeySelector struct {
  redskyappsv1alpha1.ConfigMap
  AnnotationSelector string
  LabelSelector string
  Namespace string
  CreateIfNotPresent bool
}
```

redskyctl/internal/commands/generate/experiment/configmaps.go
```diff
@@ -0,0 +1,201 @@
...
+
+// saveConfigMapKeys extracts the paths to the data values from the supplied node.
+func (r *applicationResource) saveConfigMapKeys(node *yaml.RNode, sel ConfigMapKeySelector) error {
+	// NOTE: We do not consider the `binaryData` field of the ConfigMap
+	path := []string{"data"}
+	return node.PipeE(
+		yaml.Lookup(path...),
+		yaml.FilterFunc(func(node *yaml.RNode) (*yaml.RNode, error) {
+			if mn := node.Field(sel.Key); mn != nil || sel.CreateIfNotPresent {
```
bradbeam: nit, could we drop this back a level ( think I've got the logic right on this )
```
mn := node.Field(sel.key)
if mn == nil && !sel.CreateIfNotPresent {
  return nil,nil
 }

cmv := configMapValue{}
...
```

redskyctl/internal/commands/generate/experiment/generator.go
```diff
@@ -65,6 +67,18 @@ func (g *Generator) SetDefaultSelectors() {
 			g.ReplicaSelectors[i].LabelSelector = g.Application.Parameters.Replicas.LabelSelector
 		}
 	}
+
+	if g.Application.Parameters != nil && len(g.Application.Parameters.ConfigMaps) > 0 {
```
bradbeam: nit, we should be able to drop `&& len(g.Application.Parameters.ConfigMaps) > 0`, the range below should account for it.

bradbeam (dismissed):

redskyctl/internal/commands/generate/experiment/generator.go
jgustie: ...yeah, I only put it in because of an OCD reaction to make it look like the stuff above (~where it was warranted because there is additional logic in the non-empty case~ where the thing being checked was a pointer not a list).

jgustie:
Ramped this way back. It's now just a sort order applied to patches so that config maps get patched first.

bradbeam (approved):

PR #278, opened 2021-02-11
Require a client ID and secret to reuse clients

jgustie:
Minor change to prevent clients from being reused if they do not have an ID or secret.

### DIFF
--- a/redskyctl/internal/commands/authorize_cluster/generator.go
+++ b/redskyctl/internal/commands/authorize_cluster/generator.go
@@ -214,11 +214,15 @@ func (o *GeneratorOptions) registeredClientInformation(ctx context.Context, ctrl
 		ctx = context.WithValue(ctx, oauth2.HTTPClient, oauth2.NewClient(ctx, tt.Source))
 	}
 
+	// Ignore errors or missing information and just register a new client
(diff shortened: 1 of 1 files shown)


redskyctl/internal/commands/authorize_cluster/generator.go
```diff
@@ -214,11 +214,15 @@ func (o *GeneratorOptions) registeredClientInformation(ctx context.Context, ctrl
 		ctx = context.WithValue(ctx, oauth2.HTTPClient, oauth2.NewClient(ctx, tt.Source))
 	}
 
+	// Ignore errors or missing information and just register a new client
 	info, err := registration.Read(ctx, ctrl.RegistrationClientURI, ctrl.RegistrationAccessToken)
 	if err != nil {
-		// Ignore errors and just register a new client
```
bradbeam, line 219: Is there anything helpful in `err` that we should surface?

bradbeam (approved):

redskyctl/internal/commands/authorize_cluster/generator.go
jgustie: In general, if we can't read a client back the best course of action is to just try to register another client which is exactly what returning `nil` here does. If registering the new client fails, that message will be surfaced. I also avoided `if err != nil || info.ClientID == "" || info.Client.Secret == "" { return nil }` here just for you :D

PR #281, opened 2021-02-17
feat: Use prometheus setup task in integration test

bradbeam:
This changes out the postgres experiment with a dumb experiment.
This uses the approximateRuntime to do nothing, so we dont need to
be concerned with ci runner node resources.
This also configures a prometheus setup task so we can validate/verify that
it works as expected.


hack/experiment.yaml
```diff
@@ -0,0 +1,108 @@
...
+    type: prometheus
+  trialTemplate:
+    spec:
+      approximateRuntime: 45s
+      setupServiceAccountName: redsky-setup
+      setupTasks:
+      - args:
+        - prometheus
+        - $(MODE)
+        name: monitoring
```
bradbeam: Im debating adding in a trial pod here. 
Something like containers: [ locust, nginx ] so we could add a p95 latency metric just to increase the touch points. 

Thoughts?

jgustie: I'll take you one further: should we use experiment generation instead of having the experiment committed?

jgustie: I'll take you one further: should we use experiment generation here?

hack/integration.sh
```diff
@@ -2,25 +2,36 @@
...
+
+echo "Wait for controller"
+${REDSKYCTL_BIN} check controller --wait
+
+echo "Create nginx deployment"
+kubectl apply -f hack/nginx.yaml
+
+echo "Create ci experiment"
+${REDSKYCTL_BIN} generate experiment -f hack/app.yaml > hack/experiment.yaml
+kubectl apply -f hack/experiment.yaml
```
jgustie: How about this:

```
${REDSKYCTL_BIN} generate experiment -f hack/app.yaml --include-resources | tee experiment.yaml | kubectl apply -f -
${REDSKYCTL_BIN} generate trial -f experiment.yaml --default base | kubectl create -f -
```

I don't like leaving the Git repo in a dirty state though, I wonder if `${TMPDIR}/experiment.yaml` is safe or not. Alternately, you could drop the `tee` and `generate trial` could continue using `kubectl`, just via selector: `<(kubectl get experiment -s redskyops.dev/app=ci) -o yaml` since the actual experiment name is a pain.

hack/integration.sh
```diff
@@ -2,25 +2,36 @@
 
 set -e
 
+REDSKYCTL_BIN="${REDSKYCTL_BIN:=dist/redskyctl_linux_amd64/redskyctl}"
```
jgustie, line 5: Would this be more useful for running locally if it were: `REDSKYCTL="${REDSKYCTL:-redskyctl}"` And then the CI action explicitly set the path to the full value?

hack/locustfile.py
```diff
@@ -0,0 +1,9 @@
+import time
+from locust import HttpUser, task, between
+
+class HomieUser(HttpUser):
```
jgustie, line 4: Sup

hack/nginx.yaml
```diff
@@ -0,0 +1,39 @@
...
+        resources:
+          limits:
+            memory: 25Mi
+            cpu: 50m
+          requests:
+            memory: 25Mi
+            cpu: 50m
+---
+apiVersion: v1
+kind: Service
```
jgustie: [Services first!](https://kubernetes.io/docs/concepts/cluster-administration/manage-deployment/#organizing-resource-configurations)

hack/integration.sh
bradbeam: I went this way since the main use is for CI, but allows us to do `REDSKYCTL_BIN=$(which redskyctl) hack/integration.sh` locally ( what I did to test this :D )

internal/experiment/generation/generation.go
```diff
@@ -97,10 +97,16 @@ func loadApplicationData(app *redskyappsv1alpha1.Application, src string) ([]byt
 
 	opts := []getter.ClientOption{
 		func(c *getter.Client) error {
```
bradbeam: I can put this as a separate PR if needed since its unrelated to this. But without this change, we don't work.

jgustie: This is fixed in the latest #288 changes, along with a comment clarifying why we _do not_ want use the process working directory here.

hack/integration.sh
```diff
@@ -2,25 +2,53 @@
...
+
 echo "Upload image to KinD"
 [[ -n "${IMG}" ]] && kind load docker-image "${IMG}" --name chart-testing
 [[ -n "${SETUPTOOLS_IMG}" ]] && kind load docker-image "${SETUPTOOLS_IMG}" --name chart-testing
+
 echo "Init redskyops"
-dist/redskyctl_linux_amd64/redskyctl init --wait
-echo "Create postgres experiment"
-kustomize build hack | kubectl apply -f -
+${REDSKYCTL_BIN} init
```
jgustie, line 12: Why did you remove the `--wait` here?

hack/integration.sh
```diff
@@ -2,25 +2,53 @@
...
+## Error: exit status 1
+sleep 5
+${REDSKYCTL_BIN} check controller --wait
+
+echo "Create nginx deployment"
+kubectl apply -f hack/nginx.yaml
+
+Stats() {
+  ${REDSKYCTL_BIN} generate experiment -f hack/app.yaml
+	kubectl get po -o wide
```
jgustie: Tab, also nit, since this is a script why abbreviate `pods` to `po`?

bradbeam: probably just habit, will update

hack/integration.sh
bradbeam: I can't seem to find the failed build, but I believe it was due to randomly getting an exit 1.
I'm assuming it's the same reason I had to add in the sleep below where the resources in Kubernetes weren't available/present by the time we got to waiting on the resource.

jgustie (approved):

PR #284, opened 2021-02-23
Include resources

jgustie:
Allow application resources to be emitted when generating experiments. Also, the application definition file isn't actually required.

### DIFF
--- a/redskyctl/internal/commands/generate/experiment.go
+++ b/redskyctl/internal/commands/generate/experiment.go
@@ -74,13 +74,12 @@ func NewExperimentCommand(o *ExperimentOptions) *cobra.Command {
 		RunE: commander.WithoutArgsE(o.generate),
 	}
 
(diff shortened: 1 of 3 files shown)


redskyctl/internal/commands/generate/experiment.go
```diff
@@ -166,9 +169,15 @@ func (o *ExperimentOptions) defaultNamespace() string {
 }
 
 func (o *ExperimentOptions) defaultName() string {
+	// Use the working directory
+	filename := o.Filename
+	if filename == "" || filepath.Dir(filename) == "/dev/fd" {
+		filename = "./default"
```
bradbeam: If I'm understanding the below right, should we just `return "default"` here?

jgustie: The `default` part is actually ignored. It could have been `./FUZZYBUNNY`, all we are after is the working directory name (e.g. it could have been `./app.yaml`, which illustrates why we are taking the dirname below).

bradbeam (dismissed):

bradbeam (approved):

PR #285, opened 2021-02-23
Add a linter warning when budget is too low

jgustie:


redskyctl/internal/commands/check/experiment.go
```diff
@@ -87,26 +88,32 @@ func (o *ExperimentOptions) checkExperiment(ctx context.Context) error {
 		return err
 	}
 
+	// Create a new linter for traversing the experiment and reporting errors
+	l := &linter{}
+
+	// Set the recommended budget to 20x the number of parameters
+	l.minExperimentBudget = 20 * len(exp.Spec.Parameters)
```
bradbeam, line 95: Do we need to catch >400 params?
```
if l.minExperimentBudget > 400 {
l.minExperimentBudget = 400
}
```

bradbeam (approved):

PR #287, opened 2021-03-03
Refactor scan

jgustie:
This PR serves two purposes: most importantly it moves experiment generation out of `redskyctl`. It also makes "scanning" independent of both "experiment generation" and Kustomize. A few Kustomize API references are left, but in large part the pipeline has been converted from using a Kustomize `ResMap` and API Machinery objects over to using KYAML.

The immediate benefit is that the code is significantly less complex (most everything boils down to a KYAML `Filter` function), this will make improvements to experiment generation easier. Output formatting is also improved (e.g. field ordering is more consistent). The use of KYAML pipeline also makes it easier to load resource manifests from a variety of inputs.


internal/scan/scan.go
```diff
@@ -0,0 +1,78 @@
...
+type Selector interface {
+	Select([]*yaml.RNode) ([]*yaml.RNode, error)
+	Map(node *yaml.RNode, meta yaml.ResourceMeta) ([]interface{}, error)
+}
+
+// Transformer consumes the aggregated outputs from the selectors and
+// turns them back into resource nodes. The original resource nodes used
+// to generate the inputs are also made available.
+type Transformer interface {
+	Filter(nodes []*yaml.RNode, selected []interface{}) ([]*yaml.RNode, error)
```
bradbeam, line 35: nit, should this be `Transform` instead of filter?
Thinking about reducing friction between Transformer.Filter and kio.Filter from the scanner implementation, ex
scanner.Transformer.Filter v scanner.Filter

internal/scan/runtime.go
```diff
@@ -0,0 +1,143 @@
...
+// Scheme is used globally to ensure the runtime helpers produce and consume
+// RNodes with type metadata (kind and apiVersion) and Go typing. In general,
+// there isn't a convenient way to parameterize the Read/Write calls (i.e. no
+// Context), thus the global variable. By default, the scheme is loaded with the
+// known Kubernetes API types and the types from this project (e.g Experiments).
+var Scheme = runtime.NewScheme()
+
+func init() {
+	_ = clientgoscheme.AddToScheme(Scheme)
+	_ = redskyv1beta1.AddToScheme(Scheme)
```
bradbeam, line 47: I'm assuming we can ignore redsky.v1alpha1 because the contract for the v1alpha1.application is to only generate redsky.v1beta1?

internal/scan/runtime.go
```diff
@@ -0,0 +1,143 @@
...
+	_ = redskyappsv1alpha1.AddToScheme(Scheme)
+}
+
+// ObjectSlices allows a slice of object instances to be read as resource nodes.
+type ObjectSlice []runtime.Object
```
bradbeam, line 52: should we add in the interface assertions here as well?
```
var (
  _ kio.Reader = ObjectSlice
  _ kio.Writer = ObjectList
  )
```

if it's not worth doing the above, could we add a comment to each of these types to designate the intent?

internal/scan/runtime.go
```diff
@@ -0,0 +1,143 @@
...
+
+		u := &unstructured.Unstructured{}
+		if err := u.UnmarshalJSON(data); err != nil {
+			return err
+		}
+
+		obj := newObjectForKind(u)
+		if err := Scheme.Convert(u, obj, runtime.InternalGroupVersioner); err != nil {
+			return err
+		}
```
bradbeam, line 107: does this cause an error when `obj == u` ( as in newObjectForKind did not find a match )?
trying to understand the expected behavior in this situation.

internal/scan/selector.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+
+	return true
+}
+
+// compileAnchored is a helper to make regular expressions match a full string.
+func compileAnchored(pattern string) (*regexp.Regexp, error) {
+	if pattern == "" {
+		return nil, nil
+	}
```
bradbeam, line 156: instead of returning nil here, should we do a catchall regex ( ex, `*` or `.*` )?
that should allow us to skip the `regex != nil` checks in matchesMeta

internal/scan/selector.go
```diff
@@ -0,0 +1,151 @@
...
+}
+
+// metaMatcher matches metadata. This is an alternative to `types.SelectorRegex` from Kustomize.
+type metaMatcher struct {
+	namespaceRegex *regexp.Regexp
+	nameRegex      *regexp.Regexp
+	groupRegex     *regexp.Regexp
+	versionRegex   *regexp.Regexp
+	kindRegex      *regexp.Regexp
+}
```
bradbeam, line 87: why use regex instead of strings here?

as in why wouldn't we just save off `g.Namespace` ( or `strings.Split(g.Namespace, ":")[1])` ) so we could compare `string == string` in `matchesMeta`

at the end of the day I think either implementation would work, just curious.

internal/scan/selector.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+	if m.nameRegex != nil && !m.nameRegex.MatchString(meta.Name) {
+		return false
+	}
+
+	if m.groupRegex != nil || m.versionRegex != nil {
+		group, version := "", meta.APIVersion
+		if pos := strings.Index(version, "/"); pos >= 0 {
+			group, version = version[0:pos], version[pos+1:]
+		}
```
bradbeam, line 134: should we use https://pkg.go.dev/k8s.io/apimachinery/pkg/runtime/schema#ParseGroupVersion instead?
ultimately the same, so indifferent on this suggestion

internal/scan/selector.go
```diff
@@ -0,0 +1,151 @@
...
+		return false
+	}
+	if m.nameRegex != nil && !m.nameRegex.MatchString(meta.Name) {
+		return false
+	}
+
+	if m.groupRegex != nil || m.versionRegex != nil {
+		group, version := "", meta.APIVersion
+		if pos := strings.Index(version, "/"); pos >= 0 {
+			group, version = version[0:pos], version[pos+1:]
```
bradbeam, line 133: should we switch this to `pos > 0` just to clarify if pos == 0 then the values aren't changed from the defaults above?

Or should we just start with this?
```
pos := strings.Index(meta.APIVersion, "/")
group, version = version[0:pos], version[pos+1:]
```

bradbeam (commented):
comments around internal/scan
(still need to check out internal/experiment )

internal/experiment/generator.go
```diff
@@ -0,0 +1,152 @@
...
+	redskyappsv1alpha1 "github.com/thestormforge/optimize-controller/api/apps/v1alpha1"
+	"github.com/thestormforge/optimize-controller/internal/application"
+	"github.com/thestormforge/optimize-controller/internal/experiment/generation"
+	"github.com/thestormforge/optimize-controller/internal/scan"
+	"sigs.k8s.io/kustomize/kyaml/kio"
+	"sigs.k8s.io/kustomize/kyaml/kio/filters"
+	"sigs.k8s.io/kustomize/kyaml/yaml"
+)
+
+// Generator is use to create an experiment definition.
```
bradbeam: sp `s/use/used`

internal/experiment/generator.go
```diff
@@ -0,0 +1,152 @@
...
+
+	// Always add container resource selectors, conditionally with an explicit label selector
+	var crsLabelSelector string
+	if g.Application.Parameters != nil && g.Application.Parameters.ContainerResources != nil {
+		crsLabelSelector = g.Application.Parameters.ContainerResources.LabelSelector
+	}
+	g.ContainerResourcesSelectors = []generation.ContainerResourcesSelector{
+		{
+			GenericSelector: scan.GenericSelector{
+				Group:         "apps|extensions",
```
bradbeam, line 61: Not sure I caught this in the scan implementation, but why is the format `x|y`?

internal/experiment/generator.go
```diff
@@ -0,0 +1,152 @@
...
+func (g *Generator) Execute(output kio.Writer) error {
+	return kio.Pipeline{
+		Inputs: []kio.Reader{
+			g.Application.Resources,
+		},
+		Filters: []kio.Filter{
+			&konjure.Filter{
+				Depth:         100,
+				DefaultReader: g.DefaultReader,
+				KeepStatus:    true,
```
bradbeam, line 114: Should we set `KeepStatus: false` if we're going to strip it below?

internal/experiment/generator.go
```diff
@@ -0,0 +1,152 @@
...
+		Transformer: &generation.Transformer{
+			DefaultExperimentName:       application.ExperimentName(&g.Application),
+			MergeGenerated:              len(g.Application.Scenarios) > 1,
+			IncludeApplicationResources: g.IncludeApplicationResources,
+		},
+	}
+
+	for i := range g.ContainerResourcesSelectors {
+		scanner.Selectors = append(scanner.Selectors, &g.ContainerResourcesSelectors[i])
+	}
```
bradbeam, line 138: should we skip the range and do 
```
scanner.Selectors = append(scanner.Selectors, &g.ContainerResourcesSelectors...)
```

internal/scan/runtime.go
jgustie: Originally I only had `ObjectSlice` using this scheme so it made sense to leave it off (i.e. you wouldn't generate content from the v1alpha1 structs). But, yeah with the `ObjectList` it makes sense to be there.

internal/scan/runtime.go
jgustie: When `obj == u` convert basically just does a shallow copy of the backing map, it's not an error. The end result is the unstructured object is wrapped in a `RawExtension` in the list (an unstructured object _is_ a `runtime.Object`). Code that does type checks on `Items[n].Object` will see `*unstructured.Unstructured` for any kind that does not have a Go type registered in the `Scheme`.

internal/scan/selector.go
jgustie: We allow things like `Group: "apps|extensions"`. This behavior was previously provided by Kustomize via its `types.Selector` which is used by the `ResMap` to select subsets of resources.

internal/scan/selector.go
jgustie: It just seemed like unnecessary extra use of regular expression matching in a loop. Maybe I'm just dating myself.

internal/experiment/generator.go
jgustie: This is the use of the regexp, in this case we are matching two groups from a single selector (in the previous version of the code, I had neglected to leverage the regular expression matching and had just added the selector twice: once for `apps` and once for `extensions`).

internal/experiment/generator.go
jgustie: Keeping the status here ensures objects fetched from Kubernetes using `kubectl` retain their status so we can inspect it during the scan. This is independent from striping the status later (which would ensure the experiment itself doesn't have a status field). In the case where `IncludeApplicationResources == true`, then it would also strip the status of the objects fetched from Kube, but only after giving us the opportunity to scan them.

internal/experiment/generator.go
jgustie: No, because the elements need to have their type converted independently.

internal/scan/selector.go
jgustie: They are still changed if `pos == 0`, the leading `/` is removed, e.g. `/v1` should give `"", "v1"` not `"", "/v1"`. This is consistent with the `ParseGroupVersion` function you mention above. I'll add some tests.

internal/experiment/generator.go
bradbeam: Doesn't that get handled by the variadic parameter?

bradbeam: err, I think I got what you're saying.

internal/experiment/generation/containerresources.go
```diff
@@ -0,0 +1,190 @@
...
+// matchesContainerName checks to see if the specified container name is matched.
+func (s *ContainerResourcesSelector) matchesContainerName(name string) bool {
+	// Treat empty like ".*"
+	if s.ContainerName == "" {
+		return true
+	}
+
+	containerName, err := regexp.Compile("^(?:" + s.ContainerName + ")$")
+	if err != nil {
+		// Kustomize panics. Not sure where it validates. We will just fall back to exact match
```
bradbeam: Have any more context/deets around this? Not sure I'm following where kustomize would panic from regex.Compile :)

internal/experiment/generation/generation.go
```diff
@@ -0,0 +1,102 @@
...
+import (
+	"crypto/md5"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+
+	redskyappsv1alpha1 "github.com/thestormforge/optimize-controller/api/apps/v1alpha1"
+	redskyv1beta1 "github.com/thestormforge/optimize-controller/api/v1beta1"
+	"github.com/yujunz/go-getter"
```
bradbeam, line 30: did you intend this go-getter ( instead of hashicorp )?

internal/experiment/generation/generation.go
```diff
@@ -0,0 +1,102 @@
...
+
+func init() {
+	// Hack the sort order used by the format filter to make experiments sort more naturally
+	yaml.FieldOrder["parameters"] = 100
+	yaml.FieldOrder["metrics"] = 200
+	yaml.FieldOrder["targetRef"] = 100
+	yaml.FieldOrder["patch"] = 200
+	yaml.FieldOrder["baseline"] = 100
+	yaml.FieldOrder["min"] = 200
+	yaml.FieldOrder["max"] = 300
```
bradbeam: nit, could we group up the priorities ( 100 / 200 / 300 )

internal/experiment/generation/generation.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+// newObjectiveMetric creates a new metric for the supplied objective with most fields pre-filled.
+func newObjectiveMetric(obj *redskyappsv1alpha1.Objective, query string) redskyv1beta1.Metric {
+	defer func() { obj.Implemented = true }()
+	return redskyv1beta1.Metric{
+		Type:     redskyv1beta1.MetricPrometheus,
+		Query:    query,
+		Minimize: true,
```
bradbeam, line 61: not something for this pr, but I imagine as we add in additional objectives ( ex throughput / requests/s ) we'll want to poke at this.

internal/experiment/generation/generation.go
```diff
@@ -0,0 +1,102 @@
...
+		func(c *getter.Client) error {
+			// Try to load relative to the application definition itself
+			if path := app.Annotations["config.kubernetes.io/path"]; path != "" {
+				c.Pwd = filepath.Dir(path)
+			}
+			return nil
+		},
+	}
+
+	if err := getter.GetFile(dst, src, opts...); err != nil {
```
bradbeam, line 108: kind of surprised go-getter doesn't support an io.Writer for the destination. oh well.

internal/experiment/generation/locust.go
```diff
@@ -0,0 +1,182 @@
...
+var _ MetricSource = &LocustSource{}     // Locust specific metrics
+var _ kio.Reader = &LocustSource{}       // ConfigMap for the locustfile.py
+
+func (s *LocustSource) Update(exp *redskyv1beta1.Experiment) error {
+	if s.Scenario == nil {
+		return nil
+	}
+
+	exp.Spec.TrialTemplate.Spec.JobTemplate = &batchv1beta1.JobTemplateSpec{}
+	pod := &exp.Spec.TrialTemplate.Spec.JobTemplate.Spec.Template.Spec
```
bradbeam, line 46: 👀

internal/experiment/generation/locust.go
```diff
@@ -0,0 +1,182 @@
...
+
+		case obj.Latency != nil:
+			if l := s.locustLatency(obj.Latency.LatencyType); l != "" {
+				query := `scalar(` + l + `{job="trialRun",instance="{{ .Trial.Name }}"})`
+				result = append(result, newObjectiveMetric(obj, query))
+			}
+
+		case obj.ErrorRate != nil:
+			if obj.ErrorRate.ErrorRateType == redskyappsv1alpha1.ErrorRateRequests {
+				query := `scalar(failure_count{job="trialRun",instance="{{ .Trial.Name }}"} / request_count{job="trialRun",instance="{{ .Trial.Name }}"})`
```
bradbeam, line 122: not a today thing, but I feel like it'd be worthwhile to split these out and group them up separately so we'd have a library of prom queries.

internal/experiment/generation/locust.go
```diff
@@ -0,0 +1,182 @@
...
+
+	// TODO We need to rethink how ingress scanning works, this just preserves existing behavior
+	var ingressURL string
+	if s.Application != nil && s.Application.Ingress != nil {
+		ingressURL = s.Application.Ingress.URL
+	}
+	if ingressURL == "" {
+		return fmt.Errorf("ingress must be configured when using Locust scenarios")
+	}
+	pod.Containers[0].Env = append(pod.Containers[0].Env, corev1.EnvVar{Name: "HOST", Value: ingressURL})
```
bradbeam, line 83: is it worth moving this to `locustEnv()`?

internal/experiment/generation/stormforger.go
```diff
@@ -0,0 +1,255 @@
...
+)
+
+type StormForgerSource struct {
+	Scenario    *redskyappsv1alpha1.Scenario
+	Application *redskyappsv1alpha1.Application
+}
+
+var _ ExperimentSource = &LocustSource{} // Update trial job
+var _ MetricSource = &LocustSource{}     // StormForger specific metrics
+var _ kio.Reader = &LocustSource{}       // ConfigMap for the test case file
```
bradbeam: Should these be `StormForgerSource` ?

internal/experiment/generation/locust.go
bradbeam: Should we add a method to Experiment to make this a bit more concise ( ex, Example.TrialPodSpec() *corev1.Pod )?

internal/experiment/generation/transformer.go
```diff
@@ -0,0 +1,264 @@
...
+func (p *pnode) TargetRef() *corev1.ObjectReference {
+	return &corev1.ObjectReference{
+		APIVersion: p.meta.APIVersion,
+		Kind:       p.meta.Kind,
+		Name:       p.meta.Name,
+		Namespace:  p.meta.Namespace,
+	}
+}
+
+// parameterNamer returns a name generation function for parameters based on scan results.
```
bradbeam, line 219: how boring, where is my `whale-steals-armageddon` ? :)

internal/experiment/generator.go
jgustie: https://golang.org/doc/faq#convert_slice_of_interface

internal/experiment/generation/stormforger.go
jgustie: 🤦

internal/experiment/generation/locust.go
jgustie: I wanted to leave the ingress stuff separate for the time being, I'm not sure if this is something that will even end up in here or if there will be something like an `IngressSource` that eventually sets it. Same is true in the StormForger code.

internal/experiment/generation/generation.go
jgustie: This was intensely frustrating to me. I was sure that it was just something I was missing. But as best I can tell "go getter" is really more like "go copier".

internal/experiment/generation/generation.go
jgustie: This just ends up being the starting point, the caller can always adjust it as necessary (e.g. the `DurationMetricsSource` overwrites the `Type` to be `""`).

internal/experiment/generation/generation.go
jgustie: I grouped them by attribute set instead of priority, e.g. parameters before metrics. I'll add comments.

jgustie: I've got another idea. I'll just iterate over the type's fields and add the `json:` tag names in the order that they appear...

internal/experiment/generation/generation.go
jgustie: This is the fork of go-getter that Kustomize (previously) used so it's already in `go.sum`. We can revisit when we bump to the 4.x Kustomize API and it drops out of the transient dependencies.

internal/experiment/generation/containerresources.go
jgustie: This is an old comment I probably should have cleaned up. It's in reference to the fact that Kustomize uses `MustCompile` instead of `Compile`. I never did find the code that did the error checking to give a useful error message though.

bradbeam (approved):

PR #288, opened 2021-03-04
New application command

jgustie:
This code adds a minimal "new application" command for scaffolding an app.yaml from the CLI.

At this point, the command really just helps populating the "resources" section. I think we can improve it to so it adds YAML comments to the output and maybe a few other nice to haves. This generator does take a pass over the actual resources so we may be able to get some additional information.

### DIFF
--- a/internal/application/documentation.go
+++ b/internal/application/documentation.go
@@ -107,7 +107,7 @@ func (f *DocumentationFilter) annotateApplication(app *yaml.RNode) error {
 
 	// Each key and value are elements in the content list, iterate over even indices
 	var content []*yaml.Node
-	for i := 0; i < len(n.Content); i = i + 2 {
+	for i := 0; i < len(n.Content); i = yaml.IncrementFieldIndex(i) {
 		n.Content[i].HeadComment = headComments[n.Content[i].Value]
 		content = append(content, missingRequiredContent(n.Content[i].Value, required)...)
 		content = append(content, n.Content[i], n.Content[i+1])
--- a/internal/application/documentation.go
+++ b/internal/application/documentation.go
@@ -102,6 +102,7 @@ func (f *DocumentationFilter) annotateApplication(app *yaml.RNode) error {
 		"parameters": "resources",
 		"scenarios":  "parameters",
 		"objectives": "scenarios",
+		"":           "objectives",
 	}
 
 	// Each key and value are elements in the content list, iterate over even indices
@@ -111,7 +112,9 @@ func (f *DocumentationFilter) annotateApplication(app *yaml.RNode) error {
 		content = append(content, missingRequiredContent(n.Content[i].Value, required)...)
 		content = append(content, n.Content[i], n.Content[i+1])
 	}
-	n.Content = content
+
+	// Make sure all the required content has been produced
+	n.Content = append(content, missingRequiredContent("", required)...)
 
 	return nil
 }
--- a/internal/application/documentation.go
+++ b/internal/application/documentation.go
@@ -29,28 +29,27 @@ import (
 var headComments = map[string]string{
 
 	"resources": `
-Resources define where you application's Kubernetes resources come from. These can be URL-like
-values such as file paths, HTTP URLs, or Git repository URLs. They can also be more complex
-definitions such references to in-cluster objects or Helm charts.
+Resources define where you application's Kubernetes resources come from. These
+can be URL-like values such as file paths, HTTP URLs, or Git repository URLs.
+They can also be more complex definitions such references to in-cluster objects
+or Helm charts.
 `,
 
 	"parameters": `
 Parameters control what parts of you application will be optimized.
-
 Reference: https://docs.stormforge.io/reference/application/v1alpha1/#parameters
 `,
 
 	"scenarios": `
-Scenarios determine how you application will be put under load during optimization.
-
+Scenarios determine how you application will be put under load during
+optimization.
 Reference: https://docs.stormforge.io/reference/application/v1alpha1/#scenario
 `,
 
 	"objectives": `
-Objectives are used to define what you are trying to optimize about your application. Most
-objectives correspond to metrics observed over the course of an observation trial, for
-example: "p95-latency".
-
+Objectives are used to define what you are trying to optimize about your
+application. Most objectives correspond to metrics observed over the course of
+an observation trial, for example: "p95-latency".
 Reference: https://docs.stormforge.io/reference/application/v1alpha1/#objective
 `,
 }
@@ -108,7 +107,7 @@ func (f *DocumentationFilter) annotateApplication(app *yaml.RNode) error {
 	// Each key and value are elements in the content list, iterate over even indices
 	var content []*yaml.Node
 	for i := 0; i < len(n.Content); i = i + 2 {
-		n.Content[i].HeadComment = softWrap(headComments[n.Content[i].Value], 80)
+		n.Content[i].HeadComment = headComments[n.Content[i].Value]
 		content = append(content, missingRequiredContent(n.Content[i].Value, required)...)
 		content = append(content, n.Content[i], n.Content[i+1])
 	}
@@ -145,7 +144,7 @@ func missingRequiredContent(key string, required map[string]string) []*yaml.Node
 		Kind:        yaml.ScalarNode,
 		Tag:         yaml.NodeTagString,
 		Value:       req,
-		HeadComment: softWrap(headComments[req], 80),
+		HeadComment: headComments[req],
 	})
 
 	// Add an empty value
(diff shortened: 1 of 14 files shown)


internal/application/documentation.go
```diff
@@ -0,0 +1,200 @@
...
+}
+
+// annotateApplication documents a single application.
+func (f *DocumentationFilter) annotateApplication(app *yaml.RNode) error {
+	n := app.YNode()
+
+	// Reminder about how the file was generated
+	if len(os.Args) >= 3 && filepath.Base(os.Args[0]) == "redskyctl" {
+		n.HeadComment = strings.Join(os.Args, " ")
+	}
```
bradbeam, line 95: fancy

internal/application/documentation.go
```diff
@@ -0,0 +1,200 @@
...
+	required := map[string]string{
+		"resources":  "",
+		"parameters": "resources",
+		"scenarios":  "parameters",
+		"objectives": "scenarios",
+	}
+
+	// Each key and value are elements in the content list, iterate over even indices
+	var content []*yaml.Node
+	for i := 0; i < len(n.Content); i = i + 2 {
```
bradbeam: is this to try to ensure we only insert comments on an empty line before each key?
Should we add in a `strings.HasPrefix` or `len(n.Content[i]) == 0`?

internal/application/documentation.go
```diff
@@ -0,0 +1,200 @@
...
+
+	// Required is a map of field name to required preceding field name.
+	// This is used to ensure even empty (and otherwise omitted) fields can be
+	// included for documentation purposes.
+	required := map[string]string{
+		"resources":  "",
+		"parameters": "resources",
+		"scenarios":  "parameters",
+		"objectives": "scenarios",
+	}
```
bradbeam:
```suggestion
	required := map[string]string{
  		"resources":  "",
  		"parameters": "resources",
  		"scenarios":  "parameters",
  		"objectives": "scenarios",
  		"catchall": "objectives",
  	}
```
not that I was trying to break things, but if I run `redskyctl generate application --objectives ""`, we don't get any of the required output flagged. 
If we were to add in a catchall key with a value of `objectives`, I get the expected output
```

File: internal/application/documentation.go [hunk 4]
```diff
@@ -0,0 +1,200 @@
...
+// missingRequiredContent adds top level fields which may have been empty and therefore
+// emitted during the initial Go to YAML encoding process. The supplied map is used
+// to track which fields have already been encountered and which fields must precede
+// the current field (identified by "key"). The resulting list of nodes are suitable
+// for inclusion in the content of a mapping node.
+func missingRequiredContent(key string, required map[string]string) []*yaml.Node {
+	// As soon as we encounter a key, remove it so it does not get double added
+	for k, v := range required {
+		if v == key {
+			delete(required, k)
```
bradbeam, line 131: minor optimization, should we `break` after this?

internal/application/documentation.go
```diff
@@ -0,0 +1,200 @@
...
+	result = append(result, &yaml.Node{
+		Kind:        yaml.ScalarNode,
+		Tag:         yaml.NodeTagString,
+		Value:       req,
+		HeadComment: softWrap(headComments[req], 80),
+	})
+
+	// Add an empty value
+	switch req {
+	case "parameters":
```
bradbeam, line 155: not for this pr, but I suppose a future enhancement would be to enumerate all of the konjure.resource types here.

internal/application/documentation.go
```diff
@@ -0,0 +1,200 @@
...
+			Kind:  yaml.SequenceNode,
+			Style: yaml.FoldedStyle,
+			Tag:   yaml.NodeTagSeq,
+		})
+	}
+
+	return result
+}
+
+// softWrap is a naive wrapper that really doesn't try very hard.
```
bradbeam: :)

internal/application/documentation.go
```diff
@@ -0,0 +1,200 @@
...
+`,
+
+	"objectives": `
+Objectives are used to define what you are trying to optimize about your application. Most
+objectives correspond to metrics observed over the course of an observation trial, for
+example: "p95-latency".
+
+Reference: https://docs.stormforge.io/reference/application/v1alpha1/#objective
+`,
+}
```
bradbeam: Since we have these mostly formatted up here, we can use these as is and drop the `softWrap` function. 
(nit, I'd probably drop the empty line between the definition and reference)

I played locally and set softWrap to just return the comment (without any transformation ) -- 

before
```
# Parameters control what parts of you application will be optimized. Reference: https://docs.stormforge.io/reference/application/v1alpha1/#parameters
parameters: {}

# Scenarios determine how you application will be put under load during optimization.
# Reference: https://docs.stormforge.io/reference/application/v1alpha1/#scenario
scenarios: []

# Objectives are used to define what you are trying to optimize about your application.
# Most objectives correspond to metrics observed over the course of an observation
# trial, for example: "p95-latency". Reference: https://docs.stormforge.io/reference/application/v1alpha1/#objective
objectives:
```

after
```
# Parameters control what parts of you application will be optimized.
# Reference: https://docs.stormforge.io/reference/application/v1alpha1/#parameters
parameters: {}

# Scenarios determine how you application will be put under load during optimization.
# Reference: https://docs.stormforge.io/reference/application/v1alpha1/#scenario
scenarios: []

# Objectives are used to define what you are trying to optimize about your application. Most
# objectives correspond to metrics observed over the course of an observation trial, for
# example: "p95-latency".
# Reference: https://docs.stormforge.io/reference/application/v1alpha1/#objective
objectives:
```

internal/application/generator.go
```diff
@@ -0,0 +1,187 @@
...
+		app.Objectives = append(app.Objectives, redskyappsv1alpha1.Objective{Name: o})
+	}
+}
+
+// clean ensures that the application state is reasonable.
+func (g *Generator) clean(app *redskyappsv1alpha1.Application) error {
+	var resources []konjure.Resource
+	for _, r := range app.Resources {
+		switch {
+		case r.Resource != nil:
```
bradbeam, line 164: Will this ever be nil ?

internal/experiment/generation/generation.go
```diff
@@ -95,18 +96,22 @@ func loadApplicationData(app *redskyappsv1alpha1.Application, src string) ([]byt
...
-			}
+	var opts []getter.ClientOption
+
+	// Only set the working directory to directory of the file the app.yaml was loaded from,
+	// we MUST NOT set this to the process working directory or relative paths in the app.yaml
+	// will be dependent on what directory you run the process from. If the path annotation is
+	// not present on the application, we MUST fail to load relative paths.
+	if path := app.Annotations[kioutil.PathAnnotation]; path != "" {
+		opts = append(opts, func(c *getter.Client) error {
+			c.Pwd = filepath.Dir(path)
```
bradbeam, line 107: I know earlier we set go through the work of setting this annotation to include the trailing app.yaml filename. Any reason we go through that effort instead of just using the directory itself?

internal/scan/filter.go
```diff
@@ -0,0 +1,77 @@
...
+	// We are only handling the very specific case of `kustomize build X`
+	if len(cmd.Args) != 3 || cmd.Path != "kustomize" || cmd.Args[1] != "build" {
+		return cmd.Output()
+	}
+
+	// Restrict to disk access to be consistent with the flow when we fork
+	fs := filesys.MakeFsOnDisk()
+
+	// Create Krusty options: be sure to disable KYAML as we don't want to accidentally
+	// invoke code paths in which we have created problems due to dependencies
```
bradbeam, line 63: have any specifics around this?

internal/application/documentation.go
jgustie: Yeah, that is the way smarter thing to do.

internal/application/documentation.go
jgustie: Good catch, I used `""` as a key and added a call to `missingRequiredContent` after the for loop. I'm not sure I follow the GitHub suggestion thing above, but I didn't see the part where missing content was called with the "catchall".

internal/application/documentation.go
jgustie: For how minor it would be, I'm more comfortable with the "remove all" semantics.

internal/application/documentation.go
jgustie: Agree, I just wanted to get something in for the top level. I think the actual code to do the next level down should be separate so we can account for the non-empty cases as well.

internal/application/generator.go
jgustie: Yes, one pointer in the `konjure.Resource` should be non-nil, sometimes it will be `Resource`. However the `api.Resource` is the only one that is backed by a string array. We should have a separate conversation about Konjure expansion :D

internal/experiment/generation/generation.go
jgustie: The value of the path annotation is the path to the file that the YAML was read from, normally we don't set that (it will come from inside Konjure or inside KYAML). The other code is just using `app.yaml` as a dummy for process streams to match that convention.

internal/scan/filter.go
jgustie: I'm just trying to mitigate the potential for unexpected behavior if we aren't using the exact version of KYAML that the Kustomize API is using. My concern is that upgrading the Kustomize API might cause conflicts with the old version of controller-runtime. These kinds of problems go away if we could stay on the latest versions.

internal/application/documentation.go
bradbeam: I thought the suggestion would be good to highlight the diff; we basically add an additional kv pair to the required list with a value of 'objectives'.  ex --
```
"catchall": "objectives"
```
as we traverse through the generated yaml, the missingRequiredContent handles pruning this map, so having a `catchall` ( or whatever value that is a non top level application key should work ). Since we point to objectives, we traverse through the whole map, so this extra KV pair triggers the missingRequiredContent to fill in the rest.

bradbeam (approved):

PR #289, opened 2021-03-05
Custom application configuration

jgustie:
The primary goal of this PR is to add `custom` fields to the `Application` specification for both scenarios and objectives. This will expose more functionality of the `Experiment`. In addition, there were bugs and other clean up that became apparent while adding this feature.

The following illustrates what is being added to the `Application`:

```
scenarios:
- custom:
    # This may include the built-in Prometheus so the custom trial job can push metrics 
    pushGateway: true
    # This creates an empty pod template with just the image set
    image: foobar
    # This is a full pod template
    podTemplate:
      spec:
        containers:
        - name: testing
          image: foobar
    # These settings translate to the trial spec
    initialDelaySeconds: 15
    approximateRuntimeSeconds: 15

objectives:
- name: foobar-requests
  custom:
    prometheus:
      query: 'scalar(http_requests_total{job="foobar"}'
      url: http://prometheus.monitoring:9090
- name: sum-of-something
  custom:
    maximize: true
    datadog:
      query: 'no idea'
      aggregator: sum
```

Note that the `name` field can still be specified along side the `custom` field, as can the other objective fields like min/max.


api/apps/v1alpha1/application_defaults.go
```diff
@@ -255,6 +273,39 @@ func defaultObjectiveName(values ...string) string {
...
+		switch part {
+		case "", "or", "and", "unless", "on", "ignoring", "scalar":
+			continue
+		default:
+			name = append(name, part)
+		}
+	}
+	if n := strings.Join(name, "-"); n != "" {
+		return n
+	}
```
bradbeam: Oof. 

Using the queries listed in https://prometheus.io/docs/prometheus/latest/querying/examples/ we get -- 
https://play.golang.org/p/6yTYOB7tujz

I don't think I have a better solution unfortunately.
I thought about doing a sha/checksum of the query and using that instead, but because these are metrics it'll make their usefulness pretty much 0.  Is it worth doing that to nudge people into making pretty names for these metrics but not prohibitive for getting started?

api/apps/v1alpha1/application_types.go
```diff
@@ -121,6 +123,22 @@ type LocustScenario struct {
 	RunTime *metav1.Duration `json:"runTime,omitempty"`
 }
 
+// CustomScenario is used for advanced cases where more flexibility is required.
+type CustomScenario struct {
+	// Enables Prometheus Push Gateway support for objectives that require it.
+	// The `PUSHGATEWAY_URL` environment variable will be added to all
+	// containers when the trial job starts.
+	EnablePushGateway bool `json:"pushGateway,omitempty"`
```
bradbeam: nit, unsure if `Use` instead of `Enable` would be a more appropriate

api/v1beta1/experiment_types.go
```diff
@@ -180,10 +180,10 @@ const (
 type PatchTemplate struct {
 	// The patch type, one of: strategic|merge|json, default: strategic
 	Type PatchType `json:"type,omitempty"`
-	// A Go Template that evaluates to valid patch
-	Patch string `json:"patch"`
 	// Direct reference to the object the patch should be applied to
 	TargetRef *corev1.ObjectReference `json:"targetRef,omitempty"`
+	// A Go Template that evaluates to valid patch
+	Patch string `json:"patch"`
```
bradbeam, line 186: wonder what triggered the ordering here? 
( not a blocker, just curious )

internal/experiment/generation/custom.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+
+	if d := s.Scenario.Custom.InitialDelaySeconds; d > 0 {
+		exp.Spec.TrialTemplate.Spec.InitialDelaySeconds = d
+	}
+
+	if rt := s.Scenario.Custom.ApproximateRuntimeSeconds; rt > 0 {
+		exp.Spec.TrialTemplate.Spec.ApproximateRuntime = &metav1.Duration{Duration: time.Duration(rt) * time.Second}
+	}
+
```
bradbeam, line 58: I feel like we've got similar patterns in the trial setup job codepath. would it make sense to make these utility functions there so we could reuse the same code?

#notforthispr

internal/experiment/generation/generation.go
```diff
@@ -111,3 +123,51 @@ func loadApplicationData(app *redskyappsv1alpha1.Application, src string) ([]byt
...
 	return ioutil.ReadFile(dst)
 }
+
+var metricSelectorPattern = regexp.MustCompile(`([a-zA-Z_][a-zA-Z0-9_]*)(=|!=|=~|!~)"([a-zA-Z0-9\-|]+)"`)
+
+// convertPrometheusSelector converts a Prometheus metric selector to a Kubernetes
+// label selector. This is necessary because objectives like "Requests" define their
+// "selector" as a Prometheus selector: in order for that to work with a metric
+// with `type: kubernetes` (or `type: ""`), we must first convert it over.
+func convertPrometheusSelector(metricSelector string) (*metav1.LabelSelector, error) {
```
bradbeam, line 139: is this replacing #250?

internal/experiment/generation/generation.go
```diff
@@ -111,3 +123,51 @@ func loadApplicationData(app *redskyappsv1alpha1.Application, src string) ([]byt
...
+	if metricSelector == "" {
+		return nil, nil
+	}
+
+	labelSelector := &metav1.LabelSelector{}
+	for _, ms := range strings.Split(metricSelector, ",") {
+		parts := metricSelectorPattern.FindStringSubmatch(ms)
+		if len(parts) != 4 {
```
bradbeam, line 147: I'm only catching 3 parts for the regex submatch
- `([a-zA-Z_][a-zA-Z0-9_]*)`
- `(=|!=|=~|!~)`
- `([a-zA-Z0-9\-|]+)`

testing with 
```
ms := `key="value"`
```

api/v1beta1/experiment_types.go
jgustie: The field ordering controls the serialization ordering now, and putting the target ref above the patch looks better.

internal/experiment/generation/generation.go
jgustie: No, this is because there are two different types of selectors and we need convert between them. When you configure objectives you use the field named `metricsSelector` which corresponds to the Prometheus version of a label selector. Almost everywhere else you use a field named `labelSelector` which corresponds to a Kubernetes version of label selector. In this particular case, we are given a Prometheus label selector (which would typically just be injected into a Prometheus query) but we need to use it in a context where we are selecting Kubernetes resources (the `PodList` request to the Kube API), so we must convert the selectors (to the best of our ability, Prometheus selectors are more expressive/less efficient). The point of #250 is more to reduce the limitations we have artificially enforced on the syntax of the Prometheus label selectors.

internal/experiment/generation/generation.go
jgustie: Group 0 is the entire match.

bradbeam (approved):

PR #290, opened 2021-03-06
chore: Dont force rebuild on controller build

bradbeam:

### DIFF
--- a/.github/workflows/master.yaml
+++ b/.github/workflows/master.yaml
@@ -39,20 +39,17 @@ jobs:
         run: |
           minikube start --wait all
           minikube docker-env | sed '/^#/d;s/="\(.*\)"/=\1/g;s/export //g' >> $GITHUB_ENV
-      - name: Cache Go Modules
+      - name: Go Cache
         uses: actions/cache@v1
         with:
-          path: ~/go/pkg/mod
+          path: |
+            ~/go/pkg/mod
+            ~/.cache/go-build
+            ~/Library/Caches/go-build
+            %LocalAppData%\go-build
           key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
           restore-keys: |
             ${{ runner.os }}-go-
-      - name: Cache Go Cache
-        uses: actions/cache@v1
-        with:
-          path: ~/.cache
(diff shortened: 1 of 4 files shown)


bradbeam:
the last commit is questionable if it's needed. 
out of the past 4 runs we had 1 failure in the integration test run which isn't related to this PR.

.github/workflows/master.yaml
```diff
@@ -39,20 +39,17 @@ jobs:
...
-      - name: Cache Go Modules
+      - name: Go Cache
         uses: actions/cache@v1
         with:
-          path: ~/go/pkg/mod
+          path: |
+            ~/go/pkg/mod
+            ~/.cache/go-build
+            ~/Library/Caches/go-build
+            %LocalAppData%\go-build
```
jgustie, line 49: I wonder if we should add a step that outputs `go env GOMODCACHE` and `go env GOCACHE` and reference them here instead. It's a little confusing to see 3 directories that are just the OS variants of `GOCACHE`.

.github/workflows/master.yaml
```diff
@@ -39,20 +39,17 @@ jobs:
...
           key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
```
jgustie, line 50: The hash of `go.sum` makes sense as a cache key for `GOMODCACHE` but do want both the module and build cache keyed off it? I agree that `hashFiles('**.go')` (removed below) is effectively worthless since the number of commits not touching Go files is basically 0, but should the build cache be keyed of something like the Go version?

Makefile
```diff
@@ -85,7 +85,7 @@ generate: controller-gen conversion-gen
 
 build: manifests
 	# Build on host so we can make use of the cache
-	CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -ldflags "${LDFLAGS}" -a -o manager main.go
+	CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -ldflags "${LDFLAGS}" -o manager main.go
```
jgustie, line 88: At first this gave me pause, but then I realized we were trying to leverage the build cache in GitHub Actions (though I clearly missed the cache when I switched over to macOS builds). We also don't set this in GoReleaser for the CLI. I think I am ok with it.

PR #291, opened 2021-03-06
Go 1.16 // embed.FS upgrade

bradbeam:
This updates the controller to use go 1.16 so we can make use of the newly introduced `embed.FS` functionality.

As part of this change, the kustomize package has been updated to consume `embed.FS` and the `Assets` type and related generation code goes away.


jgustie (approved):

jgustie:
Guess I best get Go upgraded.

PR #292, opened 2021-03-09
Fix relative paths

jgustie:
The bulk of this PR is to ensure that path annotations are properly set to absolute paths on resources and that those paths are use appropriately to load relative paths encountered during generation.

### DIFF
--- a/redskyctl/internal/commander/commander.go
+++ b/redskyctl/internal/commander/commander.go
@@ -33,6 +33,9 @@ import (
 	"golang.org/x/oauth2"
 	"k8s.io/apimachinery/pkg/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+	"sigs.k8s.io/kustomize/kyaml/kio"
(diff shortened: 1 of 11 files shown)


redskyctl/internal/commander/commander.go
```diff
@@ -61,6 +64,18 @@ func (s *IOStreams) OpenFile(filename string) (io.ReadCloser, error) {
 	return os.Open(filename)
 }
 
+// YAMLWriter returns a resource node writer for the current output stream. The writer
+// is configured to strip common annotations used during pipeline processing.
+func (s *IOStreams) YAMLWriter() kio.Writer {
+	return &kio.ByteWriter{
+		Writer: s.Out,
+		ClearAnnotations: []string{
+			kioutil.PathAnnotation,
```
bradbeam, line 73: Why strip the path annotation?

jgustie: The annotations are just used for internal processing and are not meant to be included with the output.

bradbeam (approved):

PR #293, opened 2021-03-10
Add a simple test-case-file flag

jgustie:


internal/application/generator.go
```diff
@@ -186,3 +207,30 @@ func (g *Generator) clean(app *redskyappsv1alpha1.Application) error {
 
 	return nil
 }
+
+// readScenario attempts to create a scenario for the application.
+func (g *Generator) readScenario() (*redskyappsv1alpha1.Scenario, error) {
+	// Make sure the file exists
+	if _, err := os.Lstat(g.ScenarioFile); err != nil {
+		return nil, err
+	}
```
bradbeam, line 216: Think this should be fine; we just care if we can read the file and not necessarily the specifics on why we can't. Meaning something like this won't be necessary -- 
```
if err != nil {
    if os.IsNotExist(err) {
    } else {
    }
}
```

bradbeam (approved):

PR #294, opened 2021-03-10
Adjust StormForger generation to account for multiple organization authorization

jgustie:
This PR consists of a few changes/fixes:

The primary change is making it possible to source the value for `STORMFORGER_JWT` from an organization specific location (in this case, a section of the TOML file). This change is largely in anticipation of running the experiment generation in an environment where multiple organizations are in use. Since a service account is tied to the organization, we must pick a token based on the org name. Note that this only affects the fallback behavior when the credentials are not specified in the `app.yaml`.

Previous versions of experiment generation can produce a trial whose pod spec references a secret that does not exist. This would be the case if you did not have `forge` logged in and did not specify credentials in your `app.yaml`. Instead of creating the invalid spec, the new behavior is to recognize the error and fail experiment generation.

An edge case where specifying your test case file as a URL instead of a file path lead to an invalid ConfigMap definition has also been addressed (instead of `filepath.Base(testCaseFile)` we now use `testCase + ".js"` in the container).

It appears that `github.com/BurntSushi/toml` is no longer supported so I switched out the TOML parser for one that was already an indirect dependency.

### DIFF
--- a/internal/experiment/generation/stormforger.go
+++ b/internal/experiment/generation/stormforger.go
@@ -22,7 +22,7 @@ import (
 	"path/filepath"
 	"strings"
 
-	"github.com/BurntSushi/toml"
+	"github.com/pelletier/go-toml"
 	redskyappsv1alpha1 "github.com/thestormforge/optimize-controller/api/apps/v1alpha1"
 	redskyv1beta1 "github.com/thestormforge/optimize-controller/api/v1beta1"
 	"github.com/thestormforge/optimize-controller/internal/scan"
@@ -46,9 +46,14 @@ func (s *StormForgerSource) Update(exp *redskyv1beta1.Experiment) error {
 		return nil
 	}
 
-	env, err := s.stormForgerEnv()
-	if err != nil {
-		return err
+	org, tc := s.stormForgerTestCase()
+	if org == "" {
+		return fmt.Errorf("missing StormForger organization")
+	}
+
+	accessToken := s.stormForgerAccessToken(org)
+	if accessToken == nil {
+		return fmt.Errorf("missing StormForger authorization")
 	}
 
 	exp.Spec.TrialTemplate.Spec.JobTemplate = &batchv1beta1.JobTemplateSpec{}
@@ -57,7 +62,20 @@ func (s *StormForgerSource) Update(exp *redskyv1beta1.Experiment) error {
 		{
 			Name:  "stormforger",
 			Image: trialJobImage("stormforger"),
-			Env:   env,
+			Env: []corev1.EnvVar{
+				{
+					Name: "TITLE",
+					ValueFrom: &corev1.EnvVarSource{
+						FieldRef: &corev1.ObjectFieldSelector{
+							FieldPath: "metadata.name",
+						},
+					},
+				},
+				{
+					Name:  "TEST_CASE",
+					Value: fmt.Sprintf("%s/%s", org, tc),
+				},
+			},
 		},
 	}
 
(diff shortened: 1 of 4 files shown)


internal/experiment/generation/stormforger.go
```diff
@@ -97,12 +114,22 @@ func (s *StormForgerSource) Update(exp *redskyv1beta1.Experiment) error {
 		pod.Containers[0].Env = append(pod.Containers[0].Env, corev1.EnvVar{Name: "TARGET", Value: ingressURL})
 	}
 
+	// Add a reference to the access token
+	pod.Containers[0].Env = append(pod.Containers[0].Env, corev1.EnvVar{
```
bradbeam, line 118: Is there a reason to handle this outside of the env setup earlier in this function ( where we set TITLE and TEST_CASE )

internal/experiment/generation/stormforger.go
```diff
@@ -112,28 +139,25 @@ func (s *StormForgerSource) Read() ([]*yaml.RNode, error) {
...
 		cm.Name = s.stormForgerConfigMapName()
-		// TODO filepath.Base is broken here if TestCaseFile is a URL with a query parameter
-		cm.Data = map[string]string{filepath.Base(s.Scenario.StormForger.TestCaseFile): string(data)}
+		cm.Data = map[string]string{tc + ".js": string(data)}
 		result = append(result, cm)
 	}
 
 	// Include a secret with the access token, if necessary
-	if accessToken := s.stormForgerAccessToken(); accessToken != nil {
+	if accessToken := s.stormForgerAccessToken(org); accessToken != nil {
```
bradbeam, line 147: With including the secretRef above in the pod spec, don't we always want to run through this flow regardless?
Im concerned we may silently allow a broken configuration that won't be immediately obvious ( if a trial pod doesn't have the secret available, is that going to just hang indefinitely ?)

internal/experiment/generation/stormforger.go
```diff
@@ -164,56 +188,70 @@ func (s *StormForgerSource) Metrics() ([]redskyv1beta1.Metric, error) {
 	return result, nil
 }
 
+func (s *StormForgerSource) stormForgerTestCase() (org, name string) {
+	parts := strings.Split(s.Scenario.StormForger.TestCase, "/")
```
bradbeam, line 192: we do `strings.Split` because test case can either be a /path/to/file.js or http://url/to/file.js ?

internal/experiment/generation/stormforger.go
jgustie: Nope, just putting the credential var at the bottom of the list so it looks nice.

bradbeam (approved):

internal/experiment/generation/stormforger.go
jgustie: If you put an explicit secret ref in the app.yaml, the assumption is that we can't verify it's existence. For example, you could have this in your app.yaml:
```
stormforger:
  accessToken:
    secretKeyRef:
      name: so-much-secret
      key: no-peeking
```
And we would generate the trial pod blindly referencing those values. This actually closes the gap were no credential information is available: previously that would have still produce a reference even though there was no expectation of the secret being generated or instructions from the user that it pre-exists.

internal/experiment/generation/stormforger.go
jgustie: The `testCaseFile` can be a path or URL the JavaScript source, the `testCase` (what we are splitting here) is just the StormForger test case name (i.e. `<org>/<name>`). I believe there is another test case name format that involves more slashes and a UUID, but I'm not 100% sure.

PR #296, opened 2021-03-10
Fix bug ignoring missing test case file

jgustie:
Move fast and break all the things.

### DIFF
--- a/internal/application/generator.go
+++ b/internal/application/generator.go
@@ -18,7 +18,6 @@ package application
 
 import (
 	"io"
-	"os"
 	"path/filepath"
 	"time"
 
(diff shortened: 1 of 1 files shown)


internal/application/generator.go
```diff
@@ -212,7 +212,7 @@ func (g *Generator) clean(app *redskyappsv1alpha1.Application) error {
 func (g *Generator) readScenario() (*redskyappsv1alpha1.Scenario, error) {
 	// Make sure the file exists
 	if _, err := os.Lstat(g.ScenarioFile); err != nil {
-		return nil, err
+		return nil, nil
```
bradbeam: Not sure I understand the change; if the scenario file doesn't exist why wouldn't we want to fail?

jgustie: This is just a blanket "ignore everything" instead of doing something like `if g.ScenarioFile == "" { return nil, nil }` first.

bradbeam (dismissed):

jgustie:
Actually, since the value can populate something that goes into go-getter, it doesn't make sense to stat the value at all.

bradbeam (approved):

PR #302, opened 2021-03-15
Streamline export application resource loading

jgustie:
This PR eliminates a couple of extra round trips we are doing during export as part of the application refactoring.


internal/scan/runtime.go
```diff
@@ -103,10 +103,10 @@ func (o *ObjectList) Write(nodes []*yaml.RNode) error {
 
 		obj := newObjectForKind(u)
 		if err := Scheme.Convert(u, obj, runtime.InternalGroupVersioner); err != nil {
-			return err
+			o.Items = append(o.Items, runtime.RawExtension{Raw: data})
```
jgustie, line 106: Previously there was a comment about CRDs, technically this is currently only being used by resources that are part of experiment generation so it isn't a problem. But if we were to `IncludeApplicationResources` on the generator this change would ensure that custom resources would effectively pass through without conversion.

internal/scan/runtime.go
```diff
@@ -144,6 +144,7 @@ func fixConversion(node *yaml.RNode) (*yaml.RNode, error) {
 func newObjectForKind(obj runtime.Object) runtime.Object {
 	if gvks, _, _ := Scheme.ObjectKinds(obj); len(gvks) > 0 {
 		if typed, err := Scheme.New(gvks[0]); err == nil {
+			typed.GetObjectKind().SetGroupVersionKind(gvks[0])
```
jgustie, line 147: Export was doing an extra Unstructured conversion to get this information, it's easier to just add while we have it.

redskyctl/internal/commands/export/export.go
```diff
@@ -422,17 +402,19 @@ func (o *Options) generateExperiment() error {
 	}
 
 	// Load up all application resources
-	res, err := apppkg.LoadResources(o.application, onDiskFS)
-	if err != nil {
-		return err
-	}
-
-	resYaml, err := res.AsYaml()
```
jgustie, line 430: We can just use a KYAML pipeline and skip the extra round trip into a Kustomize `ResMap`.

bradbeam: I think kio pipelines have taken over your mind :)

bradbeam (commented):
I think this looks good, unsure what's going on with CI

redskyctl/internal/commands/export/export.go
jgustie: 100%

PR #305, opened 2021-03-18
Turn objectives into a collections of goals

jgustie:
This PR alters the schema of `Application` in a way that breaks existing definitions. An "objective" is changed so it consists of a list of "goals". For example:
```
objectives:
- name: p95
- name: cost
```
Becomes:
```
objectives:
- goals:
  - name: p95
  - name: cost
```
Effectively, the previous `Objective` type has been renamed `Goal` and a new `Objective` type has been introduced which only has `Name` and `Goals` fields.

This change allows objectives to be specified as a single name token rather then a list. The single name can be used as a label on the experiment and will be preserved through the Experiments API (this behavior can be seen in the changes to the `export` command). This also allows generated experiments to have arbitrary names as we no longer rely on reversing their representation (we can get the scenario and objective from the labels). This also solves the problem of generating experiment names that were too long.

Additionally, the first iteration of experiment generation assumed that the `Application` would be "filtered" so that it only included a single a scenario and the relevant objectives. This model was problematic for a number of reasons and has been addressed by explicitly passing the selected scenario and objective to the parts of the code that need it.

A bulk of the changes in this PR are related to renaming "objective" to "goal" and ensuring we have access to the scenario/objective (previously we may have only required the application and assumed it was "filtered"). The `internal/application/application.go` file has changed significantly because we no longer require the complex name generation/reversal logic.


api/apps/v1alpha1/application_defaults.go
```diff
@@ -70,25 +71,42 @@ func (in *Scenario) Default() {
...
+	}
+
+	if in.Name == "" {
+		switch len(in.Goals) {
+		case 1:
+			in.Name = in.Goals[0].Name
+		case 2:
+			in.Name = fmt.Sprintf("%s-vs-%s", in.Goals[0].Name, in.Goals[1].Name)
+		default:
+			in.Name = defaultName
```
bradbeam, line 85: 🌬️ 🪁

api/apps/v1alpha1/application_defaults.go
```diff
@@ -70,25 +71,42 @@ func (in *Scenario) Default() {
 }
 
 func (in *Objective) Default() {
+	for i := range in.Goals {
+		in.Goals[i].Default()
+	}
+
+	if in.Name == "" {
```
bradbeam, line 78: nit, could we switch this to return early if name is set?

api/apps/v1alpha1/application_defaults.go
```diff
@@ -70,25 +71,42 @@ func (in *Scenario) Default() {
...
+			in.Name = fmt.Sprintf("%s-vs-%s", in.Goals[0].Name, in.Goals[1].Name)
+		default:
+			in.Name = defaultName
+		}
+	}
+}
+
+func (in *Goal) Default() {
 	// If there is no explicit configuration, create it by parsing the name
 	if in.Name != "" && in.needsConfig() {
```
bradbeam: not really part of this PR but as I was looking through it, should we move the `in.Name != ""` to `in.needsConfig()` since this is the only function that uses it?

api/apps/v1alpha1/application_defaults.go
```diff
@@ -70,25 +71,42 @@ func (in *Scenario) Default() {
 }
 
 func (in *Objective) Default() {
```
bradbeam, line 73: nit, could we change this receiver to `o` or `obj` to differentiate it from the `Goal` methods below?

api/apps/v1alpha1/application_types.go
```diff
@@ -139,12 +139,19 @@ type CustomScenario struct {
...
 }
 
-// Objective describes the goal of the optimization in terms of specific metrics. Note that only
-// one objective configuration can be specified at a time.
+// Objective describes the goals of the optimization in terms of specific metrics.
 type Objective struct {
-	// The name of the objective. If no objective specific configuration is supplied, the name is
-	// used to derive a configuration. For example, any valid latency (prefixed or suffixed with
-	// "latency") will configure a default latency objective.
+	// The name of the objective.
```
bradbeam: should we mention if no name is specified one will be auto generated based on the goals?

api/apps/v1alpha1/well_known_labels.go
```diff
@@ -25,6 +25,9 @@ const (
 	// LabelScenario is the application scenario associated with an object.
 	LabelScenario = "redskyops.dev/scenario"
 
+	// LabelObjective is the application objective associated with an object.
+	LabelObjective = "redskyops.dev/objective"
```
bradbeam, line 29: not directly related to this PR, but should we transition this now from rexdskyops.dev to apps.stormforge.io for the application spec? I'm not sure we're using these anywhere

internal/application/application.go
```diff
@@ -17,161 +17,113 @@ limitations under the License.
...
+		names = append(names, app.Scenarios[i].Name)
+		if cleanName(app.Scenarios[i].Name) == cleanScenario {
+			if result != nil {
+				return nil, fmt.Errorf("scenario name %q is ambiguous", scenario)
 			}
-			names = append(names, app.Scenarios[i].Name)
+			result = &app.Scenarios[i]
 		}
-		return fmt.Errorf("unknown scenario '%s' (should be one of '%s')", scenario, strings.Join(names, "', '"))
+	}
```
bradbeam: thoughts?
```suggestion
if cleanName(app.Scenarios[i].Name) != cleanScenario {
  continue
}

if result != nil {
  return nil, fmt.Errorf("scenario name %q is ambiguous", scenario)
}

result = &app.Scenarios[i]
names = append(names, app.Scenarios[i].Name)
```

internal/application/application.go
```diff
@@ -184,29 +136,8 @@ func WorkingDirectory(app *redskyappsv1alpha1.Application) string {
...
-	if err != nil {
-		return nil, err
-	}
-
-	dp := provider.NewDefaultDepProvider()
-	f := resmap.NewFactory(dp.GetResourceFactory(), dp.GetConflictDetectorFactory())
-	return f.NewResMapFromBytes(buf.Bytes())
-}
-
 func cleanName(n string) string {
```
bradbeam, line 145: nit, would it make sense to make this a method off Application?

internal/application/application.go
```diff
@@ -184,29 +136,8 @@ func WorkingDirectory(app *redskyappsv1alpha1.Application) string {
...
-	return strings.Map(func(r rune) rune {
+	n = strings.Map(func(r rune) rune {
```
bradbeam, line 146: not part of this PR, but while looking through this func, is `.` really valid? I think I would have expected this to fail for dns reasons.

internal/application/application_test.go
```diff
@@ -20,130 +20,45 @@ import (
...
-					{
-						Name:     "objective",
-						Requests: &v1alpha1.RequestsObjective{},
-					},
-				},
-			},
+			applicationName: "scenario-and-objective-are-normalized",
+			scenario:        "FOO?", // These are not a valid scenario/objective names
+			objective:       "BAR!",
+			expected:        "scenario-and-objective-are-normalized-8843d7f9",
```
bradbeam, line 44: im disappointed expected is not `RAWR!` :)

internal/application/generator.go
```diff
@@ -170,8 +170,10 @@ func (g *Generator) apply(app *redskyappsv1alpha1.Application) error {
 		app.Scenarios = append(app.Scenarios, *s)
 	}
 
-	for _, o := range g.Objectives {
-		app.Objectives = append(app.Objectives, redskyappsv1alpha1.Objective{Name: o})
+	if o, err := g.readObjective(); err != nil {
+		return err
+	} else if o != nil {
+		app.Objectives = append(app.Objectives, *o)
```
bradbeam, line 176: do we catch empty objectives/scenarios elsewhere?

internal/experiment/generator.go
```diff
@@ -103,7 +109,30 @@ func (g *Generator) SetDefaultSelectors() {
...
+	}
+
+	// Compute the effective scenario, objective, and experiment names
+	scenarioName, objectiveName, experimentName := "", "", g.ExperimentName
+	if scenario != nil {
+		scenarioName = scenario.Name
+	}
+	if objective != nil {
+		objectiveName = objective.Name
+	}
```
bradbeam, line 129: It seems like we're pushing this down through a few layers instead of just failing early in GetScenario/GetObjective if these are nil. Im expecting if scenario/objective are nil we're dealing with an invalid application, so would this cause issues further down in the kio pipeline?

internal/experiment/generator.go
```diff
@@ -172,20 +193,20 @@ func (g *Generator) selectors() []scan.Selector {
...
 
-	for i := range g.Application.Objectives {
-		if !g.Application.Objectives[i].Implemented {
-			return fmt.Errorf("generated experiment cannot optimize objective: %s", g.Application.Objectives[i].Name)
+	objective, err := application.GetObjective(&g.Application, g.Objective)
+	if err != nil {
+		return err
+	}
+
+	for i := range objective.Goals {
```
bradbeam, line 211: Wont this be a sad day if objective == nil?

redskyctl/internal/commands/export/export.go
```diff
@@ -453,6 +463,8 @@ func (o *Options) getTrialDetails(ctx context.Context) (*trialDetails, error) {
 	result := &trialDetails{
 		Experiment:  experimentName.Name(),
 		Application: exp.Labels["application"],
+		Scenario:    exp.Labels["scenario"],
+		Objective:   exp.Labels["objective"],
```
bradbeam, line 467: do we need to verify these keys exist first?

api/apps/v1alpha1/well_known_labels.go
jgustie: It will be part of the great 2.0 "Red Sky" purge; I have already started using `stormforge.io` on the new application values (the `experiment.ToServer` function already strips the `redskyops.dev/` so I kept them the same for consistency sake).

internal/application/application.go
jgustie: Almost, the name appending needs to be before the continue check to ensure we get the full list for error messages.

internal/application/application.go
jgustie: True story.

internal/experiment/generator.go
jgustie: I added validation to the experiment (in the transfomer) to ensure we do not produce invalid experiments (i.e. you need a parameter and a metric) since that is what is ultimately matters (e.g. `redskyctl generate experiment` fails because there are no parameters on the generated experiment). By that argument, not having a scenario is more viable then not having an objective (since an experiment is still valid without a trial job). I would rather not force the validation here to leave the door open for things like the ability to pull metric definitions while scanning.

redskyctl/internal/commands/export/export.go
jgustie: Nope, in that case they will just be set to empty strings (I love Go's zero value maps) which will match the one and only scenario or objective. If there is no label and the application definition contains multiple choices, it will fail in get `GetScenario` or `GetObjective`.

api/apps/v1alpha1/application_defaults.go
jgustie: The receiver name matches the auto-generated code to avoid lint spam.

api/apps/v1alpha1/application_defaults.go
jgustie: Your aversion for statement nesting is matched only by my aversion to commit diffs that appear unnecessarily complex because they change indentation :)

api/apps/v1alpha1/application_defaults.go
jgustie: I kind of like the name check here: it makes it more clear that the intent of the block is to assign something based on the name. I don't like the name `needsConfig` though, I will change that.

internal/application/application.go
jgustie: Only if the function was called `ValidateCreate`, `ValidateUpdate` and `ValidateDelete` :)

But in all seriousness, that is probably a better representation of what we want here: we never want the names to be invalid to begin with.

internal/application/generator.go
jgustie: This is just in the context of generation, e.g. when you put your QA hat on and give application generation `--goals ""` (which as you also pointed out, I regressed. Again).

internal/experiment/generator.go
jgustie: A sad day indeed.

bradbeam (approved):

PR #308, opened 2021-03-19
fix: Wait until the future is now :sparkles:

bradbeam:
This fixes an issue where we attempt to query for metrics in the future
after we've offset the scrape time.

Hopefully this squashes our pesky p95 scrape issue.

### DIFF
--- a/internal/metric/metric_test.go
+++ b/internal/metric/metric_test.go
@@ -140,7 +140,6 @@ func jsonPathHttpTestServer() *httptest.Server {
 	response := map[string]int{"current_response_time_percentile_95": 5}
 	return httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		json.NewEncoder(w).Encode(response)
-		return
 	}))
 }
 
@@ -148,6 +147,5 @@ func promHttpTestServer() *httptest.Server {
 	resp := `{"status":"success","data":{"resultType":"scalar","result":[1595471900.283,"1"]}}`
 	return httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		fmt.Fprint(w, resp)
-		return
 	}))
 }
--- a/internal/metric/prometheus.go
+++ b/internal/metric/prometheus.go
@@ -71,6 +71,7 @@ func capturePrometheusMetric(ctx context.Context, log logr.Logger, m *redskyv1be
 	// If we got NaN, it might be a Pushgateway metric that won't have a value unless we query from the scrape time
 	if math.IsNaN(value) && lastScrapeTime.After(completionTime) {
 		log.Info("Retrying Prometheus query to include final scrape", "lastScrapeTime", lastScrapeTime, "completionTime", completionTime)
+
 		value, err = queryScalar(ctx, promAPI, m.Query, lastScrapeTime)
 		if err != nil {
 			return 0, 0, err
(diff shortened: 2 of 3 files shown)


bradbeam:
This is an example of the logs we'd see around this. Note the log timestamp `ts` value when compared to `lastScrapeTime`.

```
NAME                    STATUS   ASSIGNMENTS         VALUES
ci-10-user-load-p68ff   Failed   memory=50, cpu=50   metric data not available (the scalar function may have received an input vector whose size is not 1)
---
{"level":"info","ts":1616082036.0007112,"logger":"controllers.Metric","msg":"Retrying Prometheus query to include final scrape","trial":"default/ci-10-user-load-p68ff","lastScrapeTime":1616082038.3649533,"completionTime":1616082033}
{"level":"info","ts":1616082036.0145996,"logger":"controllers.Metric","msg":"Retrying Prometheus query to include final scrape","trial":"default/ci-10-user-load-p68ff","lastScrapeTime":1616082038.3649533,"completionTime":1616082033}
{"level":"info","ts":1616082036.023038,"logger":"controllers.Metric","msg":"Retrying Prometheus query to include final scrape","trial":"default/ci-10-user-load-p68ff","lastScrapeTime":1616082038.3649533,"completionTime":1616082033}
{"level":"info","ts":1616082036.0314276,"logger":"controllers.Metric","msg":"Retrying Prometheus query to include final scrape","trial":"default/ci-10-user-load-p68ff","lastScrapeTime":1616082038.3649533,"completionTime":1616082033}
{"level":"error","ts":1616082036.0327082,"logger":"controllers.Metric","msg":"Metric collection failed","trial":"default/ci-10-user-load-p68ff","address":"http://redsky-default-prometheus.default:9090/","query":"scalar(p95{job=\"trialRun\",instance=\"ci-10-user-load-p68ff\"})","completionTime":1616082033,"error":"metric data not available (the scalar function may have received an input vector whose size is not 1)"
```

internal/metric/prometheus.go
```diff
@@ -71,6 +71,11 @@ func capturePrometheusMetric(ctx context.Context, log logr.Logger, m *redskyv1be
 	// If we got NaN, it might be a Pushgateway metric that won't have a value unless we query from the scrape time
 	if math.IsNaN(value) && lastScrapeTime.After(completionTime) {
 		log.Info("Retrying Prometheus query to include final scrape", "lastScrapeTime", lastScrapeTime, "completionTime", completionTime)
+		if time.Now().Before(lastScrapeTime) {
```
bradbeam: Unsure if this is the best approach or if we should change `checkReady` to return a capture error with an offset RetryAfter?

internal/metric/prometheus.go
```diff
@@ -118,18 +121,16 @@ func checkReady(ctx context.Context, api promv1.API, t time.Time) (time.Time, er
 			}
 		}
 
-		if target.LastScrape.Before(t) {
+		// Ensure we have done an additional scrape since completion time
+		if target.LastScrape.Before(t.Add(scrapeInterval)) {
```
bradbeam, line 123: Went through a few iterations of how to best address this. This seemed to be the cleanest approach to get at what we wanted.

internal/metric/prometheus.go
```diff
@@ -97,18 +98,20 @@ func capturePrometheusMetric(ctx context.Context, log logr.Logger, m *redskyv1be
 	return value, valueError, nil
 }
 
+// Choose lower then normal default scrape parameters
+// TODO We could use `api.Config` to get the actual values (global defaults and per-target settings)
+const scrapeInterval = 5 * time.Second // Prometheus default is 1m
```
bradbeam, line 103: Moved this up to package level so we could use it for testing

internal/metric/metric_test.go
```diff
@@ -140,14 +140,12 @@ func jsonPathHttpTestServer() *httptest.Server {
 	response := map[string]int{"current_response_time_percentile_95": 5}
 	return httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		json.NewEncoder(w).Encode(response)
-		return
```
bradbeam, line 143: these were called out on a linter for being unnecessary, so I dropped them

internal/metric/prometheus.go
jgustie: Wouldn't this need to be `target.LastScrape.Add(scrapeTimeout).Before(t.Add(scrapeInterval))` to ensure we are considering the time at which the scrape completed rather then when it started? Assuming that `scapeTimeout` is a viable stand in for scrape duration (which the API doesn't expose).

internal/metric/prometheus.go
```diff
@@ -118,18 +119,16 @@ func checkReady(ctx context.Context, api promv1.API, t time.Time) (time.Time, er
...
 				Address:    target.ScrapeURL,
 				RetryAfter: scrapeInterval,
 			}
 		}
 
-		// TODO This should use `target.LastScrapeDuration` once it is available
-		if ls := target.LastScrape.Add(scrapeTimeout); ls.After(lastScrape) {
-			lastScrape = ls
-		}
+		lastScrape = target.LastScrape
```
jgustie, line 131: Same as above, `target.LastScrape` is the start time of scrape, not the finish time. This needs to be the _end time_ of the scrape which we can only approximate with the scrape timeout.

bradbeam: I was hoping we would wash over those details by including an additional scrape interval on top of completion time. As long as last scrape was after completion + scrape interval, we should be covered ( under the assumption scrape timeout must be less than scrape interval ).
So by the time we get to this point, lastScrape should be at least 1 interval past completion time.

jgustie: Yeah, just realized that was what was happening as soon as I stopped thinking about it.

jgustie (approved):

PR #309, opened 2021-03-20
Add the ability to generate trial jobs

jgustie:
This has been missing for a while because I thought it was going to be a big lift: then it occured to me we could expose the functionality trivially from `generate trial`. This will allow you to add `--job trial` or `--job create` or `--job delete` to your generate trial calls and the output will be the job itself rather than the trial. This approach has the added benefit of only generating one job at a time, for example if you wanted to force a setup task clean up (you may need to explicitly set the trial job number to match up).

### DIFF
--- a/redskyctl/internal/commands/generate/trial.go
+++ b/redskyctl/internal/commands/generate/trial.go
@@ -130,15 +130,16 @@ func (o *TrialOptions) generate() error {
 	t.Annotations = nil
 
 	// If requested, convert the trial into a job
-	var obj interface{} = t
-	if o.Job != "" {
-		obj, err = newJob(t, o.Job, o.JobTrialNumber)
-		if err != nil {
-			return err
-		}
-	}
+if o.Job == "" {
+  return o.Printer.PrintObj(t, o.Out)
+}
 
-	return o.Printer.PrintObj(obj, o.Out)
+obj, err := newJob(t, o.Job, o.JobTrialNumber)
+if err != nil {
+  return err
+}
+  		
+return o.Printer.PrintObj(obj, o.Out)
 }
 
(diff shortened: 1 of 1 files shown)


redskyctl/internal/commands/generate/trial.go
```diff
@@ -121,5 +129,40 @@ func (o *TrialOptions) generate() error {
...
+	// If requested, convert the trial into a job
+	var obj interface{} = t
+	if o.Job != "" {
+		obj, err = newJob(t, o.Job, o.JobTrialNumber)
+		if err != nil {
+			return err
+		}
+	}
+
+	return o.Printer.PrintObj(obj, o.Out)
```
bradbeam: I'd prefer to drop the interface to make it a bit more clear. I get that using interface{} helps with only needing to handle a single return, but it made me question it/reread a few times. 

```suggestion
if o.Job == "" {
  return o.Printer.PrintObj(t, o.Out)
}

obj, err := newJob(t, o.Job, o.JobTrialNumber)
if err != nil {
  return err
}
  		
return o.Printer.PrintObj(obj, o.Out)
```

redskyctl/internal/commands/generate/trial.go
```diff
@@ -121,5 +129,41 @@ func (o *TrialOptions) generate() error {
 	t.Finalizers = nil
 	t.Annotations = nil
 
-	return o.Printer.PrintObj(t, o.Out)
+	// If requested, convert the trial into a job
+if o.Job == "" {
```
bradbeam: :( that formatted poorly, otherwise lgtm

bradbeam (approved):

PR #312, opened 2021-03-23
Add a "redskyops.dev/server-sync" annotation

jgustie:
The value can be "disabled" or "delete-completed" to entirely skip server synchronization or to delete server experiments on completion respectively.

### DIFF
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -303,6 +303,46 @@ func FailExperiment(exp *redskyv1beta1.Experiment, reason string, err error) boo
 	return true
 }
 
+// IsServerSyncDisabled checks to see if server synchronization should be disabled.
+func IsServerSyncDisabled(exp *redskyv1beta1.Experiment) bool {
+	switch strings.ToLower(exp.GetAnnotations()[redskyv1beta1.AnnotationServerSync]) {
+	case "disabled", "false":
+		// DO NOT perform server synchronization
+		return true
+	default:
+		// Server sync IS NOT disabled, allow server sync
+		return false
+	}
+}
+
+// DeleteServerExperiment checks to see if the supplied experiment should be
+// deleted from the server upon completion. Normally, we do not actually want to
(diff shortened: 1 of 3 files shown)


internal/server/server.go
```diff
@@ -303,6 +303,46 @@ func FailExperiment(exp *redskyv1beta1.Experiment, reason string, err error) boo
 	return true
 }
 
+// IsServerSyncDisabled checks to see if server synchronization should be disabled.
+func IsServerSyncDisabled(exp *redskyv1beta1.Experiment) bool {
+	switch strings.ToLower(exp.GetAnnotations()[redskyv1beta1.AnnotationServerSync]) {
+	case "disabled", "false":
+		// DO NOT perform server synchronization
+		return true
```
bradbeam: `if "false" return true` :)

internal/server/server.go
```diff
@@ -303,6 +303,46 @@ func FailExperiment(exp *redskyv1beta1.Experiment, reason string, err error) boo
...
+// the server.
+func DeleteServerExperiment(exp *redskyv1beta1.Experiment) bool {
+	// As a special case, check to see if synchronization is disabled. This would
+	// be the case if someone tried disabling server synchronization mid-run,
+	// presumably with the intent of not having the server experiment at the end.
+	if IsServerSyncDisabled(exp) {
+		return true
+	}
+
+	switch strings.ToLower(exp.GetAnnotations()[redskyv1beta1.AnnotationServerSync]) {
```
bradbeam, line 333: I think it would be worth using a different annotation for this behavior so we don't overload `server-sync`

jgustie: I "overloaded" the annotation intentionally, I wanted to avoid simple boolean annotations (look where that got us in our own API). I figured it would allow us to define other behaviors in the future if necessary.

bradbeam (approved):

PR #314, opened 2021-03-24
refactor: Expose generator pipeline

bradbeam:
This refactors generator.Execute to split the pipeline generation
and the pipeline execution. This should enable us to manipulate the
pipeline while still maintaining the defaults it provides.

### DIFF
--- a/internal/experiment/generator.go
+++ b/internal/experiment/generator.go
@@ -108,14 +108,25 @@ func (g *Generator) SetDefaultSelectors() {
 
 // Execute the experiment generation pipeline, sending the results to the supplied writer.
 func (g *Generator) Execute(output kio.Writer) error {
-	scenario, err := application.GetScenario(&g.Application, g.Scenario)
+	pipeline, err := g.Pipeline()
 	if err != nil {
 		return err
 	}
 
+	pipeline.Outputs = append(pipeline.Outputs, output)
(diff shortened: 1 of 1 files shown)


bradbeam:
~~This'll need some refactoring after #313 but I think there is still value in exposing the pipeline.~~ Updated

jgustie:
I'm hesitant to publicly expose the pipeline because I feel like it could lead to inconsistencies with how the experiment is generated. I had some problems with the initial implementation when I had a thinner wrapper around the pipeline (e.g. the `export` evaluation differed from the `gen exp` evaluation, now they are forced through the same code paths).

Since the `kio.Writer` interface is taken as an input to generation, you can use the same approach the `kio.ByteWriter` uses to pre-process the resource nodes (the `ByteWriter` pre-processes by clearing annotations, etc):
```
	var w kio.Writer = ...
	(&experiment.Generator{...}).Execute(kio.WriterFunc(func(nodes []*yaml.RNode) (err error) {
		nodes, err = kio.FilterAll(yaml.SetLabel("x", "y")).Filter(nodes)
		if err != nil {
			return err
		}

		return w.Write(nodes)
	}))
```
I feel this is safer than exposing the inner ordering of the pipeline which could inadvertently impact what gets generated.

PR #315, opened 2021-03-25
Handle missing baseline in generate experiment

jgustie:
Fixes a panic when no baseline is specified in the scanned manifest.

I also put replicas ahead of container resources, it makes the patches easier to read.


internal/experiment/generation/containerresources.go
```diff
@@ -64,10 +67,8 @@ func (s *ContainerResourcesSelector) Map(node *yaml.RNode, meta yaml.ResourceMet
 				p := containerResourcesParameter{pnode{
 					meta:      meta,
 					fieldPath: append(path, "[name="+name+"]", "resources"),
+					value:     rl.Value.YNode(),
 				}}
-				if rl != nil {
-					p.value = rl.Value.YNode()
-				}
```
bradbeam, line 70: so at this point, if rl was nil and CreateIfNotPresent is true, do we end up with?
```
resources: {}
```

jgustie: Effectively, yes. Technically, I don't set the key because we don't need it.

bradbeam (approved):

PR #317, opened 2021-03-26
Switch application parameters to a list

jgustie:
This PR makes the `parameters:` section of the application a list, allowing for more flexibility with how parameters scan configurations are specified. It also adds the ability change the default set of resources to optimize for. Using the [voting-webapp](https://github.com/thestormforge/examples/blob/master/voting-webapp/experiment.yaml#L9) example:

```
parameters:
- replicas:
    selector: app=vote
- containerResources:
    selector: app=vote
- containerResources:
    selector: app=db
    resources: ["memory"]
- containerResources:
    selector: app=worker
    resources: ["cpu"]
```


bradbeam (approved):
Seeing this makes me wonder if we should expose a json/yaml path field for custom parameters. So then instead of having discrete types for replicas and resources we'd have a single type for path and range (?). I could see something like that ultimately being the engine behind resource and replicas ( we provide a friendly name/alias that translates to a path ). 
more just food for thought :)

PR #322, opened 2021-03-29
Add a debug metric command

jgustie:
This adds a hidden command that generates a Prometheus unit test based on the rendered queries from an experiment's metrics (non-Prometheus metrics are just put in a YAML comment).

### DIFF
--- a/redskyctl/internal/commands/debug/metric.go
+++ b/redskyctl/internal/commands/debug/metric.go
@@ -168,6 +168,7 @@ func parseTime(input string, defaultTime time.Time) (metav1.Time, error) {
 		return metav1.NewTime(defaultTime), nil
 	}
 
+	// Look for Unix time (either `<sec-int64>` or `<sec-int64>.<nsec-int64>`)
 	ss := regexp.MustCompile(`^([0-9]+)(?:\.([0-9]+))?$`).FindStringSubmatch(input)
 	if len(ss) == 3 {
 		sec, err := strconv.ParseInt(ss[1], 10, 64)
@@ -181,6 +182,7 @@ func parseTime(input string, defaultTime time.Time) (metav1.Time, error) {
 		return metav1.NewTime(time.Unix(sec, nsec)), nil
 	}
 
+	// Fall back to RFC 3339 time
 	t, err := time.Parse(time.RFC3339Nano, input)
 	return metav1.NewTime(t), err
 }
@@ -231,6 +233,8 @@ func (p *promTest) addHeadComment(pattern string, args ...interface{}) {
 }
 
 func (p *promTest) addTest(t *redskyv1beta1.Trial) error {
+	seriesName := fmt.Sprintf(`up{job="prometheus-pushgateway", instance="redsky-%s-prometheus:9090"}`, t.Namespace)
+	seriesValues := fmt.Sprintf(`0+0x%d`, (t.Status.CompletionTime.Sub(t.Status.StartTime.Time)/(5*time.Second))+2)
 	return p.document.PipeE(
 		yaml.Lookup("tests"),
 		yaml.Append(&yaml.Node{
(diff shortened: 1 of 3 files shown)


redskyctl/internal/commands/debug/metric.go
```diff
@@ -0,0 +1,313 @@
...
+
+	_ = cmd.MarkFlagFilename("filename", "yml", "yaml")
+	_ = cmd.MarkFlagRequired("filename")
+
+	return cmd
+}
+
+func (o *MetricQueryOptions) Debug() error {
+	// Read the experiment
+	r, err := o.IOStreams.OpenFile(o.Filename)
```
bradbeam, line 76: possible enhancement for later would be to fetch this from the running cluster

redskyctl/internal/commands/debug/metric.go
```diff
@@ -0,0 +1,313 @@
...
+	t.Status.CompletionTime = &completionTime
+	return nil
+}
+
+func parseTime(input string, defaultTime time.Time) (metav1.Time, error) {
+	if input == "" {
+		return metav1.NewTime(defaultTime), nil
+	}
+
+	ss := regexp.MustCompile(`^([0-9]+)(?:\.([0-9]+))?$`).FindStringSubmatch(input)
```
bradbeam, line 172: Can you add a comment that this catches unixtime?

bradbeam, line 172: Does this allow for a non-nano unix time? as in if I do `date +%s` can I feed that in directly?

bradbeam, line 172: and can I say that I'm just as frustrated that there is not a `time.Unix/time.UnixNano` constant for the format?

redskyctl/internal/commands/debug/metric.go
```diff
@@ -0,0 +1,313 @@
...
+
+				{Kind: yaml.ScalarNode, Value: "promql_expr_test", HeadComment: "\n"},
+				{Kind: yaml.SequenceNode},
+			},
+		}),
+	)
+}
+
+func (p *promTest) addPromqlExprTest(t *redskyv1beta1.Trial, m *redskyv1beta1.Metric, q string) error {
+	return p.document.PipeE(
```
bradbeam, line 274: I think this is good. 
We may want to do some bogus/fake data insertion which we could cheese by using the arithmetic-like expression ( I don't know the actual format, but im thinking something like `1x<duration in seconds>` so we could have a hopefully easily calculable value to validate against. 
 
```
  - input_series:
      - series: 'go_goroutines{job="prometheus", instance="localhost:9090"}'
        values: '10+10x2 30+20x5' # 10 20 30 30 50 70 90 110 130
      - series: 'go_goroutines{job="node_exporter", instance="localhost:9100"}'
        values: '10+10x7 10+30x4' # 10 20 30 40 50 60 70 80 10 40 70 100 130
```

redskyctl/internal/commands/debug/metric.go
```diff
@@ -0,0 +1,313 @@
...
+
+		// Render the query (record errors as a comment)
+		q, _, err := eng.RenderMetricQueries(m, t, target)
+		if err != nil {
+			pt.addErrorComment(m, err)
+			continue
+		}
+
+		// Not Prometheus, just add a comment to the header
+		if m.Type != redskyv1beta1.MetricPrometheus {
```
bradbeam, line 130: hadn't thought about this, good catch.

bradbeam (commented):
looks good; just some questions/comments around time. because time is hard.

redskyctl/internal/commands/debug/metric.go
jgustie: Process substitution all the way. I was playing around with this using mostly `redskyctl debug metric -f <(redskyctl gen exp -f app.yaml)` but you can use `kubectl get exp foo -o yaml` also.

redskyctl/internal/commands/debug/metric.go
jgustie: I was equally surprised.

bradbeam (approved):

PR #328, opened 2021-04-06
Text/Terminal UI for running experiments

jgustie:
This remains a WIP.


redskyctl/internal/commands/run/run.go
```diff
@@ -0,0 +1,368 @@
...
+		o.lastErr = msg
+		return o, tea.Quit
+
+	}
+
+	// If the user hit esc and might quit, don't bother updating the rest of the model
+	if o.maybeQuit {
+		return o, nil
+	}
+
```
bradbeam, line 208: verifying I understand the flow on this correctly -
so user hits `esc`, maybe quit == true, this condition is met, so we return. following update call we go into the confirmation prompt to determine if it should actually quit or not?

redskyctl/internal/commands/run/run.go
```diff
@@ -0,0 +1,368 @@
...
+	meta.SetMetaDataAnnotation(&o.Generator.Application.ObjectMeta, kioutil.PathAnnotation, path)
+	return nil
+}
+
+// applyGeneratorModel takes all of the what is on the generatorModel and applies
+// it to the application in the experiment generator.
+func (o *Options) applyGeneratorModel() {
+	if o.generatorModel.NamespaceInput.Enabled() {
+
+		// TODO DEMO ONLY HACK
```
bradbeam: This still relevant ?

redskyctl/internal/commands/run/out/out.go
```diff
@@ -0,0 +1,248 @@
...
+	Ready:        {Prefix: "🎉  "},
+	Starting:     {Prefix: "🚢  "},
+	Running:      {Prefix: "👍  "},
+	Watching:     {Prefix: "👓  "},
+	Completed:    {Prefix: "🍾  "},
+	Failure:      {Prefix: "❌  "},
+	Instructions: {Style: instructionsStyle},
+	YesNo:        {OmitNewline: true},
+	Preview:      {},
+}
```
bradbeam, line 77: Todays mood is...

redskyctl/internal/commands/run/multichoiceinput/multichoiceinput.go
```diff
@@ -0,0 +1,206 @@
...
+	case tea.KeyMsg:
+		if m.Focused() {
+			switch msg.String() {
+			case "up":
+				m.Highlight(m.highlighted - 1)
+
+			case "down":
+				m.Highlight(m.highlighted + 1)
+
+			case "x":
```
bradbeam: may be worth adding the space key to this as well

redskyctl/internal/commands/run/command.go
```diff
@@ -0,0 +1,377 @@
...
+		return err
+	}
+	o.previewModel.Preview.SetContent(content)
+	return nil
+}
+
+func (o *Options) refreshTrials() tea.Msg {
+	ctx := context.TODO()
+
+	// TODO Where do we get the namespace/selector from?
```
bradbeam: Is this todo still a concern?

redskyctl/internal/commands/run/form/form.go
```diff
@@ -0,0 +1,184 @@
...
+	case tea.KeyMsg:
+		if focused != nil {
+			switch msg.Type {
+			case tea.KeyEnter:
+				// Validate the focused field when the user hits "enter"
+				// NOTE: A field like the textfield can also use `Validate` to effectively
+				// "block" a submission if there are pending actions that require "enter".
+				cmds = append(cmds, focused.Validate())
+
+			case tea.KeyPgUp:
```
bradbeam, line 87: backspace key might be worth adding here as well.

redskyctl/internal/commands/run/view.go
```diff
@@ -0,0 +1,420 @@
...
+// are in a valid state prior to starting.
+func (o *Options) initializeModel() {
+	var opts []out.FieldOption
+	if o.Verbose {
+		opts = append(opts, out.VerbosePrompts)
+	}
+
+	o.generatorModel.ScenarioType = out.FormField{
+		Prompt:       "Where do you want to get your load test from?",
+		Instructions: []string{"up/down: select", "pgup: back", "enter: continue"},
```
bradbeam, line 64: nit/dry, we could probably move some of these instructions to a variable. not a blocker but I could see this being a chore to update in the future.

redskyctl/internal/commands/run/view.go
```diff
@@ -0,0 +1,420 @@
...
+	}.NewChoiceField(opts...)
+
+	o.generatorModel.StormForgerGettingStarted = out.FormField{
+		Prompt: `Check this out to see how you can get set up with a StormForge load test:
+https://docs.stormforger.com/guides/getting-started/`,
+	}.NewExitField(opts...)
+
+	o.generatorModel.LocustfileInput = out.FormField{
+		Prompt:          "Please input a path to your Locust load test to optimize:",
+		Placeholder:     "( e.g. ~/my-project/tests/locustfile.py )",
```
bradbeam, line 85: does this need to be an absolute path or will relative work?

redskyctl/internal/commands/run/run.go
jgustie: Pretty much yes: following the first update that hits this return, we hit the `View` which sees `maybeQuit == true` and renders the prompt. The handling of the input is done above like you said. The event loop is basically: poll for msg, call `Update`, call `View`, repeat as fast as you can. The `tea.Cmd` returned by `Update` is run in a Go routine and the resulting `tea.Msg` ends up in the msg channel.

redskyctl/internal/commands/run/view.go
jgustie: Either or. We set the application path in `ReadApplication` in `run.go` to the wd so it will resolve against there.

redskyctl/internal/commands/run/form/form.go
jgustie: Backspace is tough because it might be handled by a text input as an actual backspace; we would need to special case based on the field (i.e. it would only really work for non-editable choice/multichoice fields).

redskyctl/internal/commands/run/view.go
jgustie: Yeah, furthermore pgup and enter are constant for all of them.

redskyctl/internal/commands/run/command.go
jgustie: Yes, I'll update the comment so it makes sense to humans who aren't me.

bradbeam (approved):

PR #329, opened 2021-04-07
feat: Add rhel base images

bradbeam:
This enables us to build a Certified Image for RH Partner connect.

We still need to figure out how/where to push these images,
most likely through https://access.redhat.com/RegistryAuthentication


Dockerfile.rhel
```diff
@@ -0,0 +1,19 @@
...
+
+# These labels are required for the partner guide
+# https://redhat-connect.gitbook.io/partner-guide-for-red-hat-openshift-and-container/program-on-boarding/technical-prerequisites#dockerfile-requirements
+LABEL name="Optimize" \
+      maintainer="support@stormforge.io" \
+      vendor="Stormforge" \
+      version="latest" \
+      release="1" \
+      summary="Stormforge Optimize uses machine learning to improve the performance of your application" \
+      description="Stormforge Optimize uses machine learning to improve the performance of your application."
```
jgustie: StormFrog. I mean. StormForge.

jgustie:
It seems like the Docker images are mostly duplicates of the originals which sets off my DRY alarms.

Is this something we can do with:
```
ARG BASE_IMAGE=gcr.io/distroless/static:nonroot
FROM ${BASE_IMAGE}
...
```
And then pass `--build-arg BASE_IMAGE=registry.access.redhat.com/ubi8-minimal` as a build arg from the `Makefile`? It seems like the only other changes are the labels, which we could also add with more `--label` args (I might prefer the OCI labels on the standard image so that's an argument for putting them in the `Makefile` instead of the `Dockerfile`).

Dockerfile
```diff
@@ -1,8 +1,29 @@
...
 WORKDIR /
 COPY ./manager .
-USER nonroot:nonroot
 
 ENTRYPOINT ["/manager"]
+
+# === Distroless
+FROM  base as distroless
+
+USER nonroot:nonroot
```
jgustie: I'm not aware of any requirement to use `nonroot`, would it be easier if we always just ran as `nobody`?

jgustie:
Closing this and moving the Red Hat/OpenShift container certification to a separate repository; same as #366.

PR #331, opened 2021-04-16
Add a message when there are no results for a readiness check

jgustie:
This change makes it a bit more clear when a readiness check fails because we were not able to find any resources.

### DIFF
--- a/controllers/ready_controller.go
+++ b/controllers/ready_controller.go
@@ -18,6 +18,7 @@ package controllers
 
 import (
 	"context"
+	"strings"
 	"time"
 
 	"github.com/go-logr/logr"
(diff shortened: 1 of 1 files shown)


controllers/ready_controller.go
```diff
@@ -322,6 +323,25 @@ func (rc *readinessChecker) check(ctx context.Context, c *redskyv1beta1.Readines
...
+		if c.Selector != nil {
+			missingTargetMsg.WriteString("; labels=")
+			missingTargetMsg.WriteString(metav1.FormatLabelSelector(c.Selector))
+		}
+		msg = missingTargetMsg.String()
+	}
+
 	// If a check is missing it's kind, just mark it as completed
 	if c.TargetRef.Kind == "" {
 		ok = true
```
bradbeam: Does this cause a gap?
If no items and targetRef.kind is empty, we set ok and discard `msg` ( in the if statement below, couldn't highlight the whole block).

jgustie: Would it be more clear if I moved the message construction after the potential return below?

bradbeam (dismissed):

bradbeam (approved):

PR #332, opened 2021-04-29
Minor improvements

jgustie:
This PR includes a few minor improvements:
1. Generated experiments produce parameters in the correct order, this prevents the first trial from looking different in the the `kubectl get trials` output.
2. When using `--include-resources`, the application resources are emitted first: this is more consistent with the kustomize output in the examples.
3. When using auto-complete, we no longer re-suggest duplicate values.

The first two are very minor, but the third one needed a little refactoring to get all the information in the right places.

### DIFF
--- a/redskyctl/internal/commands/experiments/arguments.go
+++ b/redskyctl/internal/commands/experiments/arguments.go
@@ -21,7 +21,6 @@ import (
 	"errors"
 	"fmt"
 	"sort"
-	"strconv"
 	"strings"
 
 	"github.com/spf13/cobra"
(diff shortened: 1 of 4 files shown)


redskyctl/internal/commands/experiments/arguments.go
```diff
@@ -42,7 +41,7 @@ func Completion(ctx context.Context, api experimentsv1alpha1.API, args []string,
 		return s, cobra.ShellCompDirectiveNoFileComp
 	}
 
-	// Assume we can parse now
+	// Assume we can names parse now
```
bradbeam, line 44: should this be assume we can parse names now?

bradbeam (approved):

redskyctl/internal/commands/experiments/arguments.go
jgustie: Names parse, we must.

PR #333, opened 2021-04-29
fix: Address minor typos in generated application

bradbeam:
- Change millisecond value to be 1s and not 10s
- Change test case name to be lowercase so the generated configmap is valid
- Add `ingress` to generated application
- Add `stormforger` to generated application
- Switch `namespaceSelector` to `namespaces` for default usage
- Switch `cost-gcp` metric to be `cost` ( we don't do anything special for one versus the other, so let's remove the option ) 

### DIFF
--- a/internal/application/documentation.go
+++ b/internal/application/documentation.go
@@ -57,6 +57,12 @@ Objectives correspond to metrics observed over the course of a trial,
 for example: "p95-latency".
 Reference: https://docs.stormforge.io/reference/application/v1alpha1/#objective
 `,
+
+		"ingress": `
+Ingress defines the destination of the load test. This is typically a public
+facing URL for your application.
+Reference: https://docs.stormforge.io/reference/application/v1alpha1/#ingress
+`,
 	}
 
 	footComments = map[string]string{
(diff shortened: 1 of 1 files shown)


internal/application/documentation.go
```diff
@@ -148,11 +167,13 @@ func (f *DocumentationFilter) annotateApplication(app *yaml.RNode) error {
...
-		"parameters": "resources",
-		"scenarios":  "parameters",
-		"objectives": "scenarios",
-		"":           "objectives",
+		"resources":   "",
+		"parameters":  "resources",
+		"scenarios":   "parameters",
+		"objectives":  "scenarios",
+		"stormForger": "objectives",
+		"ingress":     "stormForger",
```
jgustie: This should stay consistent with the declaration order in the `Application` struct since that is what the formatter will use to order the YAML fields.

jgustie (dismissed):

internal/application/documentation.go
bradbeam: As in https://github.com/thestormforge/optimize-controller/blob/master/api/apps/v1alpha1/application_types.go#L36-L53 ?

jgustie: By way of the completely non-obvious [YAML field ordering initializer](https://github.com/thestormforge/optimize-controller/blob/master/internal/experiment/generation/generation.go#L58).

jgustie (approved):

PR #335, opened 2021-05-06
TUI improvements

jgustie:
This PR contains a bunch of first round fixes to the TUI and gives more flexibility with how to handle the generated experiment (e.g. instead of hitting ctrl+t to preview, there is menu that asks if you want to run, preview or save the experiment.

### DIFF
--- a/redskyctl/internal/commands/run/out/out.go
+++ b/redskyctl/internal/commands/run/out/out.go
@@ -248,17 +248,19 @@ func (f *FormField) loadingMessage() string {
 	return " " + f.LoadingMessage + " ..."
 }
 
-// RenderKeyBindings generates an instruction line for the supplied key bindings.
-func RenderKeyBindings(keys []KeyBinding, width int) string {
-	keyStyle := instructionsStyle.Reverse()
-	descStyle := instructionsStyle
+// PagerInstructions returns a nano-style instructions bar for the pager.
+func PagerInstructions(keys []KeyBinding) func(width int) string {
+	return func(width int) string {
+		keyStyle := instructionsStyle.Reverse()
+		descStyle := instructionsStyle
 
-	var buf strings.Builder
-	buf.WriteString("\n\n")
-	tw := tabwriter.NewWriter(&buf, width/len(keys), 1, 1, ' ', 0)
-	for _, k := range keys {
-		_, _ = fmt.Fprintf(tw, "%s: %s\t", keyStyle.Styled(k.Key.String()), descStyle.Styled(k.Desc))
+		var buf strings.Builder
+		buf.WriteString("\n\n")
+		tw := tabwriter.NewWriter(&buf, width/len(keys), 1, 1, ' ', 0)
+		for _, k := range keys {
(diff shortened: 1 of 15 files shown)


redskyctl/internal/commands/run/out/out.go
```diff
@@ -224,25 +248,19 @@ func (f *FormField) loadingMessage() string {
...
-func RenderKeyBindings(keys []KeyBinding, width int) string {
-	keyStyle := instructionsStyle.Reverse()
-	descStyle := instructionsStyle
-
-	var buf strings.Builder
-	buf.WriteString("\n\n")
-	tw := tabwriter.NewWriter(&buf, width/len(keys), 1, 1, ' ', 0)
-	for _, k := range keys {
-		_, _ = fmt.Fprintf(tw, "%s: %s\t", keyStyle.Styled(k.Key.String()), descStyle.Styled(k.Desc))
+// PagerInstructions returns a nano-style instructions bar for the pager.
```
bradbeam, line 251: >  nano-style 

say it aint so :)

redskyctl/internal/commands/run/pager/pager.go
```diff
@@ -0,0 +1,99 @@
...
+	)
+
+	switch msg := msg.(type) {
+
+	case tea.WindowSizeMsg:
+		m.Model.Width = msg.Width
+		m.Model.Height = msg.Height
+		if m.Instructions != nil {
+			m.instructions = m.Instructions(m.Model.Width)
+			m.instructions = strings.Replace(m.instructions, "\r\n", "\n", -1)
```
bradbeam, line 56: Not sure I noticed the line endings before. Any additional information on where this is coming from/why we need to convert?

redskyctl/internal/commands/run/view.go
```diff
@@ -129,26 +138,33 @@ https://docs.stormforger.com/guides/getting-started/`,
 	}
 
 	o.generatorModel.ContainerResourcesSelectorInput = out.FormField{
-		Prompt:       "Specify labels to control discovery of memory and CPU parameters:",
-		Placeholder:  "( e.g. component=api )",
-		Instructions: []string{"Leave blank to select all resources", "pgup: back", "enter: continue"},
+		Prompt:      "Specify labels to control discovery of memory and CPU parameters:",
+		Placeholder: "( e.g. component=api )",
+		Instructions: []interface{}{
```
bradbeam, line 143: Seems like the whitespace alignment is off/changed? Just to double check gofmt was run on this right?

bradbeam (approved):

redskyctl/internal/commands/run/pager/pager.go
jgustie: Really I'm just mimicking what the bubbletea view port is doing to prevent any strange edge cases. I don't think it's actually necessary (all the instruction text in our case comes from `out.go`), it's just there for consistency.

redskyctl/internal/commands/run/view.go
jgustie: The array changed from a single line to a multi-line declaration so the alignment reset.

PR #336, opened 2021-05-11
Fix command

jgustie:
This PR adds a `redskyctl fix -f experiment.yaml` command. The goal of this command is to provide something to aid in the eventual migration away from the `redskyops.dev` group name. The current implementation handles all of the v1alpha1 -> v1beta1 Experiment object migrations.

Note that this migration path is one way (sequenced as migrations from version N-1 to N) and is performed directly on the YAML (not the Go structures). It does not handle migration of trial data. This tool is intended to be used on stored experiment files outside of the cluster. Although it solves a different problem, when we move away from redskyops.dev the current scheme based migration code will become obsolete and will be removed (i.e. this will be the only way to get from a v1alpha1 to where ever we end up).

### DIFF
--- a/internal/sfio/migrate.go
+++ b/internal/sfio/migrate.go
@@ -1,0 +1,212 @@
+/*
+Copyright 2021 GramLabs, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package sfio
+
+import (
+	"net/url"
+	"strings"
+
+	"github.com/thestormforge/konjure/pkg/filters"
+	redskyv1alpha1 "github.com/thestormforge/optimize-controller/api/v1alpha1"
+	redskyv1beta1 "github.com/thestormforge/optimize-controller/api/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"sigs.k8s.io/kustomize/kyaml/yaml"
+)
+
+// ExperimentMigrationFilter is a KYAML filter for perform experiment migration.
+type ExperimentMigrationFilter struct {
+}
+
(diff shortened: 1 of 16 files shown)


internal/sfio/migrate.go
```diff
@@ -0,0 +1,212 @@
...
+type metric struct {
+	Type     string                `yaml:"type"`
+	Scheme   string                `yaml:"scheme"`
+	Selector *metav1.LabelSelector `yaml:"selector"`
+	Port     intstr.IntOrString    `yaml:"port"`
+	Path     string                `yaml:"path"`
+}
+
+// setURLField returns a filter that will set the named field with the metric URL.
+func (m *metric) setURLField(name string) yaml.Filter {
```
bradbeam, line 177: is there any of this that we can reuse/migrate/refactor from https://github.com/thestormforge/optimize-controller/blob/master/api/v1alpha1/experiment_conversion.go#L105

redskyctl/internal/commander/commander.go
```diff
@@ -64,6 +65,51 @@ func (s *IOStreams) OpenFile(filename string) (io.ReadCloser, error) {
...
+		return &kio.ByteReader{
+			Reader:         s.In,
+			SetAnnotations: map[string]string{kioutil.PathAnnotation: path},
+		}
+	}
+
+	return &kio.ByteReader{
+		Reader:         &fileReader{Filename: filename},
+		SetAnnotations: map[string]string{kioutil.PathAnnotation: filename},
+	}
```
bradbeam: thoughts on reducing the returns?

```suggestion
	var reader io.Reader = &fileReader{Filename: filename}
	annotations := map[string]string{kioutil.PathAnnotation: filename}

	if filename == "-" {
  		path := "/dev/stdin"
  		if namedReader, ok := s.In.(interface{ Name() string }); ok {
  			path = namedReader.Name()
  		}
  		reader = s.In
  		annotations[kioutil.PathAnnotation] = path
  	}
 
  	return &kio.ByteReader{
  		Reader:         reader,
  		SetAnnotations: annotations,
  	}
```

bradbeam (commented):
looks good, just a couple minor comments

internal/sfio/migrate.go
jgustie: I mentioned it in the description, that code is going away (along with the rest of the `v1alpha1` package). Once this PR is in, I'll cut a v2 branch and start poking at those changes (I want this fix command in both 1.x and 2.x though).

redskyctl/internal/commander/commander.go
jgustie: Actually, in thinking about it, we don't want the path set to `/dev/stdin` because then it will try to resolve relative file paths against `/dev/`.

bradbeam (approved):

PR #337, opened 2021-05-12
Add lint warning when SMP is used on a non-core type

jgustie:
This change adds a warning when using a strategic merge patch on a kind that is not recognized by client-go.


bradbeam (approved):
🔥

PR #341, opened 2021-05-20
SFIO improvements

jgustie:
This PR cleans up some of the recent SFIO additions (comments); and adds a few bits of new functionality:
1. There is a replacement for the Kustomize `types.FieldPath` parser that also supports placeholders
2. Offers the ability to preserve the `RNode.fieldPath` through a `yaml.FieldMatcher`
3. Adds a `TeeMatchedFilter` to make `yaml.PathMatcher` usable in a pipeline (also preserving `RNode.fieldPath`)

Note that I am working on some fixes to the `ContainResourceSelector` which will be the first consumer for the new stuff, that isn't quite ready.

### DIFF
--- a/internal/experiment/generation/quantity.go
+++ b/internal/experiment/generation/quantity.go
@@ -42,30 +42,30 @@ func AsScaledInt(q resource.Quantity, scale resource.Scale) int32 {
 	return int32(v)
 }
 
-type suffixList []string
-
-func (l suffixList) lookup(scale resource.Scale) string {
-	i := int(scale)/3 + 3
-	if int(scale)%3 != 0 || i < 0 || i >= len(l) {
-		return ""
-	}
-	return l[i]
-}
-
 var (
-	binarySuffix  = suffixList{"", "", "", "", "Ki", "Mi", "Gi", "Ti", "Pi", "Ei"}
-	decimalSuffix = suffixList{"n", "u", "m", "", "k", "M", "G", "T", "P", "E"}
+	binarySuffix  = []string{"", "", "", "", "Ki", "Mi", "Gi", "Ti", "Pi", "Ei"}
+	decimalSuffix = []string{"n", "u", "m", "", "k", "M", "G", "T", "P", "E"}
 )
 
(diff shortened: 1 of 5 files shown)


internal/experiment/generation/quantity.go
```diff
@@ -0,0 +1,72 @@
...
+		for e := int(scale) / 3; e < 0; e++ {
+			v *= 1024
+		}
+	}
+	return int32(v)
+}
+
+type suffixList []string
+
+func (l suffixList) lookup(scale resource.Scale) string {
```
bradbeam: Can we add a comment to this?

Im pretty sure it's to normalize the value of scale so we can use this as the index for the suffixLists below, but not fully understanding what resource.Scale ( https://pkg.go.dev/k8s.io/apimachinery/pkg/api/resource#Scale ) was ahead of time took me a bit to work through this.

internal/sfio/fieldpath_test.go
```diff
@@ -0,0 +1,85 @@
...
+		path     string
+		data     map[string]string
+		expected []string
+	}{
+		{
+			desc: "empty",
+		},
+		{
+			desc: "tricky empty",
+			path: `/{ .ItsATrap }`,
```
bradbeam, line 37: :ItsATrap:!!!

internal/sfio/fieldpath_test.go
```diff
@@ -0,0 +1,85 @@
...
+		},
+		{
+			desc:     "run of leading slashes",
+			path:     `//spec/replicas`,
+			expected: []string{"spec", "replicas"},
+		},
+		{
+			desc:     "no data",
+			path:     `/spec/{ .WhichWayDidHeGo }`,
+			expected: []string{"spec", ""},
```
bradbeam, line 69: I'm disappointed this doesn't return `[]string{"spec", "george"}` or `[]string{"spec", "HeWentThatAWay"}`

bradbeam (dismissed):

bradbeam (approved):

PR #344, opened 2021-05-26
Allow metrics to select information from the trial job

jgustie:
This PR allows metrics to query the trial job in a similar to fashion to what we support for patches. This is useful because the trial job may have state (such as individual container timings) that are relevant as metrics. The change uses the same `IsTrialJobReference` test that the patching code uses.


parsenz (approved):
Reads like it says on the tin! Looks good to me.

PR #346, opened 2021-05-27
Environment variable scanning

jgustie:
This PR extends the app.yaml syntax to support discovery and tuning of environment variables. Like container resource scanning, the current implementation only exposes the functionality for deployments and stateful sets.

Environment variables can be added to the existing `parameters` section, you specify an environment variable name and the label selector matching the deployments/statefulsets you want to search on. Only environment variables with an explicit value can be tuned (i.e. no `valueFrom` support). Scanning can recognize a fixed prefix and/or suffix on the value: for example if an environment variable is defined as `value: '--mem 128mb'` you can supply a prefix of `'--mem '` and a suffix of `'mb'` to isolate and optimize on the numeric value `128`. Finally, you can specify a list categorical values to optimize over (the baseline will be added to this list implicitly and the prefix and suffix are still allowed).

Examples:
```
parameters:
# Look for the FOO_BAR environment variable and strip the "x" suffix
- environmentVariable:
    name: FOO_BAR
    suffix: x
# Optimize ALPHABET over the values "a", "b", "c"
- environmentVariable:
    name: ALPHABET
    values:
    - a
    - b
    - c
# Only consider deployments/stateful sets with `app=env` labels
- environmentVariable:
    name: NOT_ALL_RESOURCES
    selector: app=env
```

### DIFF
--- a/internal/experiment/generation/environmentvariables.go
+++ b/internal/experiment/generation/environmentvariables.go
@@ -1,0 +1,145 @@
+/*
+Copyright 2021 GramLabs, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package generation
+
+import (
+	"fmt"
+	"regexp"
+	"strconv"
+	"strings"
+
+	redskyv1beta1 "github.com/thestormforge/optimize-controller/api/v1beta1"
+	"github.com/thestormforge/optimize-controller/internal/scan"
+	"github.com/thestormforge/optimize-controller/internal/sfio"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"sigs.k8s.io/kustomize/kyaml/yaml"
+)
+
+// EnvironmentVariablesSelector scans for environment variables.
+type EnvironmentVariablesSelector struct {
+	scan.GenericSelector
+	// Regular expression matching the container name.
+	ContainerName string `json:"containerName,omitempty"`
+	// Name of the environment variable to match.
+	VariableName string `json:"variableName,omitempty"`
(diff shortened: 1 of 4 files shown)


internal/experiment/generation/environmentvariables.go
```diff
@@ -0,0 +1,145 @@
...
+	// Name of the environment variable to match.
+	VariableName string `json:"variableName,omitempty"`
+	// Path to the environment variable's value.
+	Path string `json:"path,omitempty"`
+	// Prefix that appears before the value.
+	ValuePrefix string `json:"valuePrefix,omitempty"`
+	// Suffix that appears after the value.
+	ValueSuffix string `json:"valueSuffix,omitempty"`
+	// Allowed values for categorical parameters.
+	Values []string `json:"values,omitempty"`
```
bradbeam, line 46: Should we embed the EnvironmentVariable struct in here?

internal/experiment/generation/transformer.go
```diff
@@ -262,21 +262,36 @@ func parameterNamer(selected []interface{}) ParameterNamer {
...
+
+		if name != "" {
+			parts = append(parts, name)
+		}
+
+		// Explainer: Parameter names are used in Go Templates which are executed
+		// against Go structs, if the template parser encounters a token that is
+		// not a valid Go field name, parsing fails (e.g. "bad character U+002D '-'").
+
+		parameterName := strings.Join(parts, "_")
```
bradbeam, line 292: do we need to worry about parameterName length <=63 chars ?

bradbeam (approved):

internal/experiment/generation/transformer.go
jgustie: The parameter name only lives in our API space, so it shouldn't need to be limited at all (it is just persisted on the backend as JSON). The only reason we need to worry about the "_" is the Go text/template parser and I think we can get around that restriction in our own template Engine by passing maps instead of structs as the `data` to the text/template `Execute` function (which is just kind of a pain). I ran some quick tests and something like `{{.Values.foo-bar}}` works fine if `data` is a `map[string]interface{}` but fails if `data` is a `struct` (and I suspect that `data` cannot be a `struct` with a `map` field, since that is our current implementation in the template Engine).

internal/experiment/generation/environmentvariables.go
jgustie: I'm sticking with the approach from the other implementations for now, I would like to switch them all over at once when we expose more of the internals in the app.yaml API.

PR #349, opened 2021-05-28
Update the Optimize Common library to pickup a round of StormForge updates

jgustie:
There will be another release of optimize-go to pick up the latest refactorings (i.e. the rename of the configuration), but we aren't even on the current release of optimize-go yet.

Also note that I am targeting a long running v2 branch for this work. There are a lot of changes that are needed for the renaming so I'm going to break them up so we can just merge the v2 branch all at once.

### DIFF
--- a/redskyctl/internal/commands/export/export_api_test.go
+++ b/redskyctl/internal/commands/export/export_api_test.go
@@ -106,7 +106,11 @@ func (f *fakeRedSkyServer) GetExperiment(ctx context.Context, name string) (expe
 	return experimentsapi.Experiment{}, nil
 }
 
-func (f *fakeRedSkyServer) CreateExperiment(ctx context.Context, name experimentsapi.ExperimentName, exp experimentsapi.Experiment) (experimentsapi.Experiment, error) {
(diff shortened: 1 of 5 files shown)


redskyctl/internal/commands/export/export_api_test.go
```diff
@@ -106,7 +106,11 @@ func (f *fakeRedSkyServer) GetExperiment(ctx context.Context, name string) (expe
 	return experimentsapi.Experiment{}, nil
 }
 
-func (f *fakeRedSkyServer) CreateExperiment(ctx context.Context, name experimentsapi.ExperimentName, exp experimentsapi.Experiment) (experimentsapi.Experiment, error) {
+func (f *fakeRedSkyServer) CreateExperimentByName(ctx context.Context, name experimentsapi.ExperimentName, experiment experimentsapi.Experiment) (experimentsapi.Experiment, error) {
```
parsenz, line 109: Does the `fakeRedSkyServer` here need to be updated to `fakeOptimizeServer` or similar?

parsenz (approved):

redskyctl/internal/commands/export/export_api_test.go
jgustie: I'll get that with the 0.0.10 update when I'm doing the bulk of the type renaming.

PR #352, opened 2021-05-28
Update Import aliases

jgustie:
This PR updates the import aliases to remove "redsky" references. There was some inconsistencies with import alias usage that was normalized as well.

This one was painful but it accounts for the largest number of "Red Sky" textual matches in the project.


parsenz (approved):
😌  That was a lot! Looks good. Thanks!

PR #354, opened 2021-06-01
Update conditions to stormforge.io

jgustie:
This PR updates the condition values (those used in trial and experiment status, and those used in readiness gates) to reflect the new domain prefix.

In working on the migration code, I discovered a couple more edge case bugs in the `TeeMatcher` which have been resolved. It turns out the `PathMatcher` also allows for matching scalar values in which case the match is not recorded (this was used to match the `conditionTypes` field and uncovered an off-by-one bug).


parsenz (approved):
Looks good! Thanks.

PR #358, opened 2021-06-01
Drop support for v1alpha1

jgustie:
This PR removes the entire `v1alpha1` implementation and all of it's conversion. This means the controller will return to supporting a single version of the CRD only. This removal is relatively superficial: a bulk of the `runtime.Scheme` base conversion we do in the code is no longer necessary but would take significantly more effort to remove, e.g. we could simplify things like YAML reading/writing by using KYAML instead of the Kubernetes runtime.

In several cases, the `LegacyHostnamePlaceholder` was replaced by a string literal: this is only intended to be temporary as we will completely drop support for pre-URL metrics in v2; that will happen after the API version is renamed.


controllers/metric_controller.go
```diff
@@ -337,7 +336,7 @@ func (r *MetricReconciler) resolveLegacyURL(ctx context.Context, t *optimizev1be
 	}
 
 	// Look for the special placeholder hostname that indicates we should look up a service
-	if u.Hostname() != optimizev1alpha1.LegacyHostnamePlaceholder {
+	if u.Hostname() != "redskyops.dev" {
```
parsenz, line 339: Does it make sense to introduce a constant for this so changing it to something else down the road is a lot easier - perhaps just in the one spot?

parsenz (approved):
Looks good - just have the one suggestion.

controllers/metric_controller.go
jgustie: I mentioned it in the description, the use of redskyops.dev as a placeholder is going away, I really just need something more obvious to make sure I can grep out where it is still being used.

PR #359, opened 2021-06-01
Pin yaml.v3 due to incompatible formatting changes

jgustie:
This PR pins the version of the YAML library to match the latest version supported by KYAML. This is necessary because the YAML project changed their formatting in way that does not match the Kubernetes conventions. The YAML API remains compatible so pinning should be safe (this is the same thing KYAML is doing).


parsenz:
Looks like tests are unhappy here.

jgustie:
The worst is when you make a change to the logs to get more information and that "fixes" the problem...

parsenz (approved):
Looks good. Thanks.

PR #360, opened 2021-06-03
Update optimize-go

jgustie:
This PR updates to the latest optimize-go which forces two changes:
1. The configuration object is renamed from `RedSkyConfig` to `OptimizeConfig`
2. The namespace and prefix are changed from `redsky-system` and `redsky-` to `stormforge-system` and `optimize-`


hack/integration.sh
```diff
@@ -22,10 +22,10 @@ stats() {
   ${REDSKYCTL_BIN} generate experiment -f hack/app.yaml
   echo "::endgroup::"
   echo "::group::Describe Application Resources"
-  kubectl describe trials,jobs,pods -l redskyops.dev/application=ci
+  kubectl describe trials,jobs,pods -l stormforge.io/application=ci
```
jgustie, line 25: I rebased `part-deux` onto `master` and picked up this mismatched label that needed to be corrected.

parsenz (approved):

PR #364, opened 2021-06-08
Remove legacy selector to host migration

jgustie:
This PR removes the runtime component of the v1alpha1 to v1beta1 migration and adjusts the offline migration logic accordingly.

The backstory is that in v1alpha1 metrics were configured to use a `Service` selector: when it was time to collect metrics we would search for matching services in the cluster and build an appropriate URL. In v1beta1 we switched to having metrics configured using a URL. To ease the transition, we injected a placeholder host name (`redskyops.dev`) to v1beta1 metrics that triggered a subset of the original v1alpha1 behavior (i.e. if the URL host name is "redskyops.dev" then look up the `Service` using the selector instead). With v1alpha1 gone, we no longer need to worry about that transition so it can be removed; any experiments which were relying on the transition will simply fail to migrate.

### DIFF
--- a/controllers/metric_controller.go
+++ b/controllers/metric_controller.go
@@ -20,7 +20,6 @@ import (
 	"context"
 	"fmt"
 	"math"
-	"net/url"
 	"strconv"
 
 	"github.com/go-logr/logr"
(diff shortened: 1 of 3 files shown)


jameshochadel (approved):
Makes sense. Do we have a new migration path for the alpha resources, or are they simply not supported in the new major controller version?

jgustie:
They are not supported at all (all the code was removed); since the group changed they were going to be different resources, not different versions of the same resources so it made sense to drop them now. The `redskyctl fix` command will take the v1alpha1 redskyops.dev/Experiments and make them into the v1beta2 optimize.stormforge.io/Experiments.

PR #366, opened 2021-06-10
Add a Red Hat UBI based image

jgustie:
This replaces #329 with a slightly different approach; we only update the main controller image and there are a few more tweaks after reviewing submission guidelines (e.g. the license). Other minor differences are that I added OCI labels to the default image and the equivalent Red Hat labels on the UBI image and adjusted the run IDs in the manifests. This PR also includes publication to Red Hat's registry.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,2 +1,3 @@
 **
 !/manager
+!/LICENSE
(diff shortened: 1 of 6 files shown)


jgustie:
I think we need to take a broader look at how we are building and pushing images.

jgustie:
Closing this and moving the Red Hat/OpenShift container certification to a separate repository; many of the changes here have been incorporated into the mainline already via #367 and #383.

PR #367, opened 2021-06-10
Clean up

jgustie:
This PR address a number of miscellaneous issues spotted during the 2.0 work:
1. There were a number of Kustomize usages in different contexts that are no longer necessary
2. The `results` command should have been deleted two minor versions ago (in 1.10)
3. We are only including a single OCI label on all our Docker images

With the OCI labels, I did not include all the labels on all the images: I didn't want to be hard coding unique information all over the place (I also did not want to require an overwhelming number of environment variables to do a build); and we are likely going to break this repository up so only a single Docker image is being produced.

With Kustomize, a few things changed: supporting the plugin model didn't make any sense because it literally has never been used. We also changed how we generate manifests when Go 1.16 introduced `embed` and we were left with a few remnants. Finally, we used to require `kustomize` for our integration tests, but now that functionality is handled by the experiment generation.


Makefile
```diff
@@ -1,4 +1,11 @@
 
+# Project metadata
+TITLE := Optimize Controller
+VENDOR := StormForge
+EMAIL := techsuport@stormforge.io
```
jameshochadel:
```suggestion
EMAIL := techsupport@stormforge.io
```

jameshochadel (approved):

PR #379, opened 2021-07-12
refactor: Move api client creation to server package

bradbeam:
This moves our api client functions to the server package. This should set the
stage for adding in the application api client code since they both should
share a very similiar initialization.

Note, this requires the latest optimize-go code, so we can't merge until a new optimize-go release is cut.

### DIFF
--- a/internal/server/client.go
+++ b/internal/server/client.go
@@ -1,0 +1,82 @@
+/*
+Copyright 2020 GramLabs, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


internal/server/client.go
```diff
@@ -0,0 +1,82 @@
...
+
+	"github.com/thestormforge/optimize-controller/v2/internal/version"
+	"github.com/thestormforge/optimize-go/pkg/api"
+	experimentsv1alpha1 "github.com/thestormforge/optimize-go/pkg/api/experiments/v1alpha1"
+	"github.com/thestormforge/optimize-go/pkg/config"
+)
+
+const audience = "https://api.carbonrelay.io/v1/"
+
+func NewExperimentClient(ctx context.Context, uaComment string) (experimentsv1alpha1.API, error) {
```
jgustie: nit: Can we call this `NewExperimentAPI` to match how it's used?

jgustie (approved):

PR #381, opened 2021-07-16
feat: Add support for in cluster generation

bradbeam:
This also includes #382 ( we can either keep included with this PR and close the other or keep separate )

### Reminders
- ~Ensure we cut a new release of Konjure~
- ~Ensure we cut a new release of Optimize-go~


internal/experiment/runner.go
```diff
@@ -0,0 +1,291 @@
...
+
+			if err := r.apiClient.DeleteActivity(ctx, activity.URL); err != nil {
+				r.handleErrors(ctx, fmt.Errorf("%s: %w", "failed to delete activity", err), activity.URL)
+			}
+		}
+	}
+}
+
+func (r *Runner) handleErrors(ctx context.Context, err error, activityURL string) {
+	failure := applications.ActivityFailure{FailureMessage: err.Error()}
```
jgustie, line 200: This should have a reason code: generally the error checking looks for a non-empty reason code but doesn't look at the message.

internal/experiment/runner.go
```diff
@@ -0,0 +1,291 @@
...
+
+	activityCh := make(chan applications.ActivityItem)
+	go subscriber.Subscribe(ctx, activityCh)
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case activity := <-activityCh:
+			// TODO might want to consider moving this to a func so we can defer delete activity
```
jgustie: I like this idea.

internal/experiment/runner.go
jgustie: This should have a reason code: generally the error checking looks for a non-empty reason code but doesn't look at the message.

internal/experiment/runner.go
```diff
@@ -0,0 +1,291 @@
...
+}
+
+// This doesnt necessarily need to live here, but seemed to make sense
+func (r *Runner) Run(ctx context.Context) {
+	if r.apiClient == nil {
+		return
+	}
+
+	query := applications.ActivityFeedQuery{}
+	query.SetType("poll", applications.TagScan, applications.TagRun)
```
jgustie: `"poll"` shouldn't be here, it isn't a tag (it's a made up hub type).

internal/experiment/runner.go
```diff
@@ -0,0 +1,291 @@
...
+	}
+
+	return &Runner{
+		client:    kclient,
+		apiClient: appAPI,
+		log:       logger,
+	}, nil
+}
+
+// This doesnt necessarily need to live here, but seemed to make sense
```
jgustie: This makes golint sad.

main.go
```diff
@@ -135,6 +137,15 @@ func main() {
...
 	// +kubebuilder:scaffold:builder
 
+	runner, err := experiment.New(mgr.GetClient(), ctrl.Log.WithName("generation").WithName("experiment"))
+	if err != nil {
+		setupLog.Error(err, "unable to start applicaiton service runner", "application", "runner")
+		os.Exit(1)
+	}
+
+	ctx := context.Background()
+	go runner.Run(ctx)
```
jgustie: This means the runner never gets to shutdown cleanly, I wonder if we should be leveraging `mgr.Add(...)` to get the manager's shutdown message:
```
// Alternately, there is a `manager.RunnableFunc`...
func (r *Runner) Start(stop <-chan struct{}) error {
	ctx, cancel := context.WithCancel(context.Background())
	go func() {
		<-stop
		cancel()
	}()
	r.Run(ctx)
	return nil
}

// ...then you can do
mgr.Add(runner)
```

controllers/server_controller.go
```diff
@@ -73,6 +74,7 @@ type ServerReconciler struct {
 	Log            logr.Logger
 	Scheme         *runtime.Scheme
 	ExperimentsAPI experimentsv1alpha1.API
+	// ApplicationsAPI applicationsv1alpha1.API
```
jgustie: ?

internal/experiment/scan.go
```diff
@@ -0,0 +1,201 @@
...
+
+package experiment
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+
+	"github.com/thestormforge/konjure/pkg/konjure"
+	redskyappsv1alpha1 "github.com/thestormforge/optimize-controller/v2/api/apps/v1alpha1"
```
jgustie: This should be `optimizeappsv1alpha1`.

internal/experiment/scan.go
```diff
@@ -0,0 +1,201 @@
...
+	if err := validate(app, scenario); err != nil {
+		return nil, err
+	}
+
+	// TODO this might(?) belong in internal/server
+
+	// Construct a controller representation of an application from the api definition
+	baseApp := &redskyappsv1alpha1.Application{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: scenario.Name,
```
jgustie: Shouldn't this use `app.Name`?

internal/experiment/scan.go
```diff
@@ -0,0 +1,201 @@
...
+
+		if res.Kubernetes.Namespace == "" {
+			return nil, fmt.Errorf("invalid resources, no namespace specified")
+		}
+
+		kResources = append(kResources, res)
+	}
+
+	return kResources, nil
+}
```
jgustie: I think this can be simplified to just:
```
func (r *Runner) scanResources(app applications.Application, baseApp optimizeappsv1alpha1.Application) error {
  data, err := json.Marshal(app.Resources)
  if err != nil {
    return err
  }
  return json.Unmarshal(data, &baseApp.Resources)
}
```
We do not need to enforce the resource type (even if the application API won't ever give us anything else), and the validation constraints here are more limiting then what is allowed by Konjure (even if the application API can't represent it).

internal/experiment/scan.go
```diff
@@ -0,0 +1,201 @@
...
+
+	g := &Generator{
+		Application: app,
+	}
+
+	// Exposed for testing so we can pass through
+	// fake kubectl output
+	if r.kubectlExecFn != nil {
+		g.FilterOptions = scan.FilterOptions{KubectlExecutor: r.kubectlExecFn}
+	}
```
jgustie: This can just be `g.FilterOptions.KubectlExecutor = r.kubectlExecFn` (embrace the default values!)

internal/experiment/scan.go
```diff
@@ -0,0 +1,201 @@
...
+
+	sf := redskyappsv1alpha1.StormForgerScenario{}
+	if err := json.Unmarshal(rawSF, &sf); err != nil {
+		return nil, err
+	}
+
+	appScenario := []redskyappsv1alpha1.Scenario{
+		{
+			// TODO Should this be name or title?
```
jgustie: Name is correct, we don't have human readable names in `optimizeappsv1alpha1`. But to that point, you should be able to round trip a whole scenario:
```
func (r *Runner) scanScenarios(scn applications.Scenario, baseApp optimizeappsv1alpha1.Application) error {
  data, err := json.Marshal(scn)
  if err != nil {
    return err
  }
  baseApp.Scenarios = make([]optimizeappsv1alpha1.Scenario, 1)
  return json.Unmarshal(data, &baseApp.Scenarios[0])
```
Though this might depend on OPT-767 to make sure all the names are in alignment re: branding.

internal/server/api.go
```diff
@@ -19,64 +19,79 @@ package server
...
 	// Get the Experiments API endpoint from the configuration
 	// NOTE: The current version of the configuration has an explicit configuration for the
 	// experiments endpoint which would duplicate the "/experiments/" path segment
 	srv, err := config.CurrentServer(cfg.Reader())
 	if err != nil {
-		return "", nil, err
+		return nil, err
 	}
 
 	address := strings.TrimSuffix(srv.API.ExperimentsEndpoint, "/v1/experiments/")
```
jgustie: This is a dumb problem to have, but the configuration model and the API client model diverged when the API version became part of the service path. The unfortunate side effect is that:
```
address := strings.TrimSuffix(srv.API.ExperimentsEndpoint, "/v1/experiments/")
```
Needs to be:
```
address := strings.TrimSuffix(srv.API.ApplicationsEndpoint, "/v2/applications/")
```
Even though they _should_ effectively produce the same value.

internal/experiment/runner.go
bradbeam: didn't know if runner made sense to keep under `internal/experiment` or move to a separate package. I think initially I had it here so I could manipulate some of the generate functionality, but I don't think that's necessary anymore.

controllers/server_controller.go
bradbeam: left over from trying to do too much at once :oops:

internal/experiment/scan.go
bradbeam: 👍 
there's a chance this was left over from before I had updated konjure and had to do a little manual intervention

cli/internal/commands/authorize_cluster/generator.go
```diff
@@ -295,7 +295,8 @@ func performanceTokens(label, controllerName string, data map[string][]byte) con
 	hasControllerEnvVar := func(name string) bool { return len(data[name]) > 0 }
 
 	// List all the organizations the user belongs to
-	orgs, err := exec.Command("forge", "organization", "list", "--output", "plain").Output()
+	// Janky hack mate
+	orgs, err := exec.Command("false", "forge", "organization", "list", "--output", "plain").Output()
```
jgustie: Should we make StormForger credentials opt-in for now?

cli/internal/commands/grant_permissions/generator.go
```diff
@@ -171,6 +175,66 @@ func (o *GeneratorOptions) generateClusterRole(roleRef *rbacv1.RoleRef) *rbacv1.
...
 
+	// :godmode:
+	// if o.CreateApplicationServicesRBAC {
+	if o.CreatePrometheusRBAC {
+		clusterRole.Rules = append(clusterRole.Rules,
+			rbacv1.PolicyRule{
+				Verbs:     []string{""},
+				APIGroups: []string{""},
+				Resources: []string{""},
+			},
```
jgustie: Was this a WIP template?

cli/internal/commands/grant_permissions/generator.go
```diff
@@ -77,6 +80,7 @@ func (o *GeneratorOptions) addFlags(cmd *cobra.Command) {
 	cmd.Flags().BoolVar(&o.CreateTrialNamespaces, "create-trial-namespace", o.CreateTrialNamespaces, "include trial namespace creation permissions")
 	cmd.Flags().StringVar(&o.NamespaceSelector, "ns-selector", o.NamespaceSelector, "bind to matching namespaces")
 	cmd.Flags().BoolVar(&o.IncludeManagerRole, "include-manager", o.IncludeManagerRole, "bind manager to matching namespaces")
+	cmd.Flags().BoolVar(&o.CreatePrometheusRBAC, "prometheus", true, "roles to create Prometheus in cluster")
```
jgustie: This should default to the current value in the struct, then to adjust the default when adding the subcommand ([here](https://github.com/thestormforge/optimize-controller/blob/4015d2b82c4acfa4c62a1e7e46df3239cbfc4dd8/cli/internal/commands/generate/generate.go#L44) and [here](https://github.com/thestormforge/optimize-controller/blob/4015d2b82c4acfa4c62a1e7e46df3239cbfc4dd8/cli/internal/commands/initialize/generator.go#L179])).

internal/server/api.go
```diff
@@ -18,64 +18,79 @@ package server
...
 	// NOTE: The current version of the configuration has an explicit configuration for the
 	// experiments endpoint which would duplicate the "/experiments/" path segment
 	srv, err := config.CurrentServer(cfg.Reader())
 	if err != nil {
-		return "", nil, err
+		return nil, err
 	}
 
-	address := strings.TrimSuffix(srv.API.ExperimentsEndpoint, "/v1/experiments/")
+	address := strings.TrimSuffix(srv.API.ApplicationsEndpoint, "/v2/applications/")
```
jgustie: This still doesn't address my original point: the way we are setup right now you should not use the same client for the application and experiment APIs. The least impactful hack around this that I can think of right now would be something like this:
```
func newClientFromConfig(ctx context.Context, uaComment string, address func(config.Server) string) (api.Client, error) {
  ...
  c, err := api.NewClient(address(srv), rt)
  ...
}

# From NewExperimentAPI call:
client, err := newClientFromConfig(ctx, uaComment, func(srv configServer) string {
  return strings.TrimSuffix(srv.API.ExperimentsEndpoint, "/v1/experiments/")
})

# From NewApplicationAPI call:
client, err := newClientFromConfig(ctx, uaComment, func(srv configServer) string {
  return strings.TrimSuffix(srv.API. ApplicationsEndpoint, "/v2/applications/")
})
```

controllers/application_poller_controller.go
```diff
@@ -0,0 +1,278 @@
...
+	return &Poller{
+		client:    kclient,
+		apiClient: appAPI,
+		log:       logger,
+	}, nil
+}
+
+func (p *Poller) Start(ch <-chan struct{}) error {
+	if p.apiClient == nil {
+		<-ch
```
jgustie: Just return immediately and let the Go routine complete?

cli/internal/commands/authorize_cluster/generator.go
bradbeam: This was a workaround(?) because I've got a service account token in my stormforge.toml that I wanted to use. I'm not sure if the service account is limited to not being able to create additional service accounts, but without triggering the err here, I was unable to get the storm forger jwt added to my controller deployment

cli/internal/commands/grant_permissions/generator.go
bradbeam: ☠️ 
yeah just a copy/paste template

internal/server/api.go
bradbeam: they're still different http clients, just both configured to the same endpoint.
but do we want to have them split? I thought since we'll be proxying requests through applications endpoint shouldn't they be using the same endpoint?

jgustie: The issue is that the client was refactored to be shared between the APIs (i.e. the "v1" was removed from the path), but the configuration was not: there are still distinct entry points for experiments and applications represented in the configuration. Technically the two trim calls above _should_ produce the same results if you are taking the defaults, but the config does expose the contract for changing them independently.

controllers/application_poller_controller.go
```diff
@@ -0,0 +1,278 @@
...
+// is converted into an api.Template consisting of parameters and metrics.
+// When an ActivityItem is tagged with run, the previous scanned template results are merged with
+// the results of an experiment generation workflow. Following this, the generated resources are applied/created
+// in the cluster.
+// note, rbac defined in cli/internal/commands/grant_permissions/generator
+func (p *Poller) handleActivity(ctx context.Context, activity applications.ActivityItem) {
+	// We always want to delete the activity after having received it
+	defer func() {
+		if err := p.apiClient.DeleteActivity(ctx, activity.URL); err != nil {
+			p.handleErrors(ctx, fmt.Errorf("%s: %w", "failed to delete activity", err), activity.URL)
```
jgustie: This should probably just log the error instead of trying to update the activity it just failed to delete.

internal/server/application.go
```diff
@@ -0,0 +1,300 @@
...
+		return nil, err
+	}
+
+	appScenario := optimizeappsv1alpha1.Scenario{}
+
+	if err = json.Unmarshal(data, &appScenario); err != nil {
+		return nil, err
+	}
+
+	// Handle internal apis referencing `stormforger` instead of `stormforge-perf`
```
jgustie: We should get the app API updated so it's returning the right thing.

controllers/application_poller_controller.go
```diff
@@ -0,0 +1,278 @@
...
+				return
+			}
+		}
+
+		p.log.Info("successfully created in cluster resources", "activity", activity.URL)
+	}
+}
+
+func (p *Poller) handleErrors(ctx context.Context, err error, activityURL string) {
+	failure := applications.ActivityFailure{FailureMessage: err.Error()}
```
jgustie: Still needs a reason or the failure never happened.

controllers/application_poller_test.go
```diff
@@ -0,0 +1,205 @@
...
+						Optimize: &pfalse,
+					},
+				},
+			},
+		},
+	}
+
+	scheme := runtime.NewScheme()
+	redskyv1beta2.AddToScheme(scheme)
+	redskyappsv1alpha1.AddToScheme(scheme)
```
jgustie: `s/redsky/optimize/g`

controllers/application_poller_controller.go
bradbeam: yeah that'd probably be fine. assuming I misinterpreted what https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.5.0/pkg/manager#RunnableFunc was meaning by `it being important that it blocks until it<I read this as manager, but probably means function> is done`

cli/internal/commands/authorize_cluster/generator.go
bradbeam: ( it shouldn't have been included here, was a sloppy git add )

controllers/application_poller_controller.go
jgustie: Yeah, the runnables just get sent off into a [Go routine](https://github.com/kubernetes-sigs/controller-runtime/blob/0fcf28efebc9a977c954f00d40af966d6a4aeae3/pkg/manager/internal.go#L225).

controllers/application_poller_controller.go
```diff
@@ -0,0 +1,288 @@
...
+			holder := &unstructured.Unstructured{}
+			holder.SetGroupVersionKind(generatedResources[i].GetObjectKind().GroupVersionKind())
+			err = p.client.Get(ctx, objKey, holder)
+			switch {
+			case apierrors.IsNotFound(err):
+				if err := p.client.Create(ctx, generatedResources[i]); err != nil {
+					p.handleErrors(ctx, fmt.Errorf("%s: %w", "failed to create object", err), activity.URL)
+					return
+				}
+			case err == nil:
```
bradbeam, line 257: lets not support updates for this round; 
log a message saying this would have triggered an update, but don't do anything

controllers/application_poller_controller.go
bradbeam: Reasons ( scan failed / run failed )

controllers/application_poller_controller.go
```diff
@@ -0,0 +1,288 @@
...
+	generatedResources, err := p.generateApp(*assembledApp)
+	if err != nil {
+		p.handleErrors(ctx, fmt.Errorf("%s (%s): %w", "failed to generate application", activity.URL, err), activity.URL)
+		return
+	}
+
+	var exp *optimizev1beta2.Experiment
+	for i := range generatedResources {
+		if expObj, ok := generatedResources[i].(*optimizev1beta2.Experiment); ok {
+			exp = expObj
```
bradbeam, line 188: Add annotations to experiment
- experiment url ( from experiments url link + generated name of experiment )
-   - api.RelationExperiments + generatedName

controllers/application_poller_controller.go
```diff
@@ -0,0 +1,288 @@
...
+	// Set defaults for application
+	app.Default()
+
+	// TODO hack from above issue ( missing namespace )
+	if app.Namespace == "" {
+		app.Namespace = "default"
+	}
+
+	g := &experiment.Generator{
+		Application: app,
```
bradbeam: set experiment name to strings.tolower(ulid.mustnew(xxx))

controllers/application_poller_controller.go
```diff
@@ -0,0 +1,308 @@
...
+
+	var exp *optimizev1beta2.Experiment
+	for i := range generatedResources {
+		if expObj, ok := generatedResources[i].(*optimizev1beta2.Experiment); ok {
+			exp = expObj
+			if exp.ObjectMeta.Annotations == nil {
+				exp.ObjectMeta.Annotations = make(map[string]string)
+			}
+
+			exp.ObjectMeta.Annotations[optimizev1beta2.AnnotationExperimentURL] = fmt.Sprintf("%s/%s", experimentURL, exp.ObjectMeta.Name)
```
jgustie: You can use the `metav1` helper here (it will allocate the missing map for you); also I think there is usually a bit of confusion over trailing slashes in URLs so it might be safer to account for both cases:

```go
metav1.SetMetaDataAnnotation(&exp.ObjectMeta, optimizev1beta2.AnnotationExperimentURL, strings.TrimRight(experimentURL, "/") + "/" + exp.Name)
```

jgustie (approved):

PR #385, opened 2021-08-02
Add missing application/scenario labels on fix

jgustie:
This PR adds a step in the `v1beta1/Experiment` to `v1beta2/Experiment` migration that ensures both the application and scenario labels are set (as they will be soon be required on all `v1beta2/Experiment` instances).

Additionally, I changed the name of the container used during experiment generation to match the scenario itself (rather then the type of scenario): for example, when generating an experiment using a StormForger test case named "jeremy/black-friday" the old behavior would have been to use "stormforger" as the container name while the new behavior will be to call that container "black-friday". This also aligns with the behavior implemented in the migration (where a missing scenario name will be inferred from the first container name when the explicit label is missing).

### DIFF
--- a/internal/sfio/migrate.go
+++ b/internal/sfio/migrate.go
@@ -297,7 +297,7 @@ func (f *ExperimentMigrationFilter) appLabels(node *yaml.RNode) (labelApplicatio
 
 		// Filter out buzzwords, and pure integers
 		var appName []string
-		drop := regexp.MustCompile("^example|experiment|final|yolo|[0-9]+$")
+		drop := regexp.MustCompile("^(example|experiment|final|yolo|[0-9]+)$")
 		for _, p := range strings.Split(strings.ToLower(md.Name), "-") {
 			if p != "" && !drop.MatchString(p) {
 				appName = append(appName, p)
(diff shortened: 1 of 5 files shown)


internal/sfio/migrate.go
```diff
@@ -262,6 +282,49 @@ func (f *ExperimentMigrationFilter) migrateMetricsV1alpha1(node *yaml.RNode) (*y
...
+	}
+
+	// Guess an application name using the experiment name
+	labelApplication = md.Labels[optimizeappsv1alpha1.LabelApplication]
+	if labelApplication == "" {
+		labelApplication = "default"
+
+		// Filter out buzzwords, and pure integers
+		var appName []string
+		drop := regexp.MustCompile("^example|experiment|final|yolo|[0-9]+$")
```
bradbeam: ❤️ 
Only filter that might be contentious would be stripping the numbers. I could imagine something like `myapp30` and `myapp40` if I were to try to associate version with the name

bradbeam (dismissed):

internal/sfio/migrate.go
jgustie: No, that's a bug (the regexp is missing parens); it should let `myapp30` through, but it should take `myapp-30` down to `myapp`.

bradbeam (approved):

PR #389, opened 2021-08-10
Replace Red Sky samples

jgustie:
There were two "Red Sky" samples that were basically left over from the initial Kubebuilder scaffolding. This PR replaces them with `optimize.stormforge.io` versions that contain at least a few populated fields.

The main reason for doing this now is to fulfil an [OperatorHub requirement](https://k8s-operatorhub.github.io/community-operators/packaging-required-fields/): the `alm-examples` annotation contains a JSON list of sample resources for each of the CRDs owned by the operator. For an Operator SDK bootstrapped project, this annotation is populated when generating the operator bundle using the contents of the `/config/samples` directory (via inclusion in the Kustomization piped into `operator-sdk generate bundle`).

These samples are shown on the OperatorHub generated preview (for example, you can see the presentation [here](https://operatorhub.io/operator/elastic-cloud-eck) at the bottom of the page). Because they are converted to JSON to be stored in the annotation, I have omitted what might have been otherwise useful documentation comments in the samples.

### DIFF
--- a/config/samples/optimize.stormforge.io_v1beta2_experiment.yaml
+++ b/config/samples/optimize.stormforge.io_v1beta2_experiment.yaml
@@ -1,0 +1,60 @@
+apiVersion: optimize.stormforge.io/v1beta2
+kind: Experiment
+metadata:
+  name: experiment-sample
+  labels:
+    stormforge.io/application: my-app
+    stormforge.io/scenario: black-friday
+spec:
+  parameters:
+  - name: cpu
+    baseline: 1000
+    min: 500
+    max: 2000
+  - name: memory
+    baseline: 2048
+    min: 1024
+    max: 4096
+  metrics:
+  - name: duration
+    minimize: true
+    query: "{{ duration .StartTime .CompletionTime }}"
+  - name: resource_requests
+    minimize: true
+    query: '{{ resourceRequests .Target "cpu=0.017,memory=0.000000000003" }}'
+    target:
+      apiVersion: v1
+      kind: PodList
+  patches:
+  - targetRef:
+      name: my-app
+      apiVersion: apps/v1
+      kind: Deployment
+    patch: |
+      spec:
+        template:
+          spec:
+            containers:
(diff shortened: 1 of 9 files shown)


jgustie:
Two late additions:
1. The samples are now embedded into the binary and can be emitted using `stormforge generate sample --kind Experiment`
2. Application files (like tests) and application resources (via the Konjure update) can now be inlined: any multiple line input will be treated as file content. If you need to specify a file name that contains a newline, it will need to be done via a `file:` URL with percent encoding (and hopefully go-getter is good with that).

config/samples/optimize.stormforge.io_v1beta2_trial.yaml
```diff
@@ -0,0 +1,39 @@
...
+            - name: LOAD_TEST_URL
+              value: http://my-app/
+  assignments:
+  - name: cpu
+    value: 1000
+  - name: memory
+    value: 2048
+status:
+  phase: Created
+  assignments: 'cpu=1000, memory=2048'
```
bradbeam, line 39: Should we drop the status field here?

config/samples/optimize.stormforge.io_v1beta2_experiment.yaml
```diff
@@ -0,0 +1,60 @@
+apiVersion: optimize.stormforge.io/v1beta2
+kind: Experiment
+metadata:
```
bradbeam, line 3: does this need meta.annotations.documentation block?

bradbeam (approved):

jgustie:
I put the "documentation" annotation in and left the status on the trial to try and convey the non-user-editable nature of the trial (under "normal" circumstances). I do not plan to include the `kind: Trial` example in the Cluster Service Version (part of the reason there is a `--kind` flag on the `gen sample`). I found an [example](https://operatorhub.io/operator/argocd-operator) of a listing on Operator Hub that does this and it just omits the "View YAML Example" link.

PR #396, opened 2021-08-19
Improve StormForge Performance integration

jgustie:


internal/experiment/generation/stormforgeperfaz.go
```diff
@@ -0,0 +1,203 @@
...
+
+func (az *StormForgePerformanceAuthorization) accessTokenForOrganization(ctx context.Context, org string) (string, error) {
+	// Check for an organization specific environment variable (this will never conflict with "real forge")
+	key := fmt.Sprintf("STORMFORGER_%s_JWT", strings.Map(toEnv, org))
+	if at, ok := az.lookupEnv(key); ok {
+		return at, nil
+	}
+
+	// Register a new service account
+	sa, err := exec.CommandContext(ctx, "forge", "serviceaccount", "create", org, az.ServiceAccountLabel()).Output()
```
bradbeam, line 73: One thing to note is this requires the user to be a manager/owner of the current org in forge. Im not sure what the error message looks like if don't have those permissions, but it may be worth enriching the error message to mention this.

bradbeam (dismissed):

bradbeam (approved):

PR #401, opened 2021-08-23
fix: Try to use namespace from kubernetes resource for in cluster generation

bradbeam:
This attempts to get the namespace from the application resources to use
as the namespace for experiments.

### DIFF
--- a/controllers/application_poller_controller.go
+++ b/controllers/application_poller_controller.go
@@ -176,6 +176,22 @@ func (p *Poller) handleActivity(ctx context.Context, activity applications.Activ
 		return
 	}
 
+	// Use resource namespaces for application namespace.
+	// defaulting to `default`
+	assembledApp.Namespace = "default"
+
+	if len(assembledApp.Resources) == 1 && assembledApp.Resources[0].Kubernetes != nil {
+		switch {
+		case assembledApp.Resources[0].Kubernetes.Namespace != "":
+			assembledApp.Namespace = assembledApp.Resources[0].Kubernetes.Namespace
+		case len(assembledApp.Resources[0].Kubernetes.Namespaces) > 0:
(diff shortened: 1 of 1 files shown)


controllers/application_poller_controller.go
```diff
@@ -176,6 +176,22 @@ func (p *Poller) handleActivity(ctx context.Context, activity applications.Activ
...
+		switch {
+		case assembledApp.Resources[0].Kubernetes.Namespace != "":
+			assembledApp.Namespace = assembledApp.Resources[0].Kubernetes.Namespace
+		case len(assembledApp.Resources[0].Kubernetes.Namespaces) > 0:
+			assembledApp.Namespace = assembledApp.Resources[0].Kubernetes.Namespaces[0]
+		}
+	} else {
+		p.handleErrors(ctx, log, activity.URL, ActivityReasonGenerationFailed, "No kubernetes resources defined", nil)
+		return
+	}
```
jgustie: This feels overly constrained, we should just take the first Kubernetes resource and let generation decide if there are no resources defined (i.e. don't fail here, wait for generation to return the error):
```
for i := range assembedApp.Resources {
  k := assembedApp.Resources[i].Kubernetes
  if  k == nil {
    continue
  }

  // Guess the namespace based on the Kubernetes resource
  // NOTE: We are not trying to evaluate the namespace selector here, this may need to go through the loop again
  if k.Namespace != "" {
    assembledApp.Namespace = k.Namespace
    break
  }
  if len(k.Namespaces) > 0 && k.Namespaces[0] != "" {
    assembledApp.Namespace = k.Namespaces[0]
    break
  }
}
```

jgustie: An alternate to this approach might be to include resources in generation and adjust the experiment namespace after the fact, however this is more complex and more expensive to compute. Nevermind.

jgustie (approved):

PR #402, opened 2021-08-24
feat: Refactor Poller.Start to use updated subscriber logic 

bradbeam:
This refactor makes use of the updated subscriber.Subscribe logic to be blocking.
This should address the issues we saw while using for/select over a closed channel.

### DIFF
--- a/controllers/application_poller_controller.go
+++ b/controllers/application_poller_controller.go
@@ -19,6 +19,7 @@ package controllers
 import (
 	"context"
 	"crypto/rand"
+	"errors"
 	"fmt"
 	"os/exec"
 	"strings"
@@ -79,7 +80,14 @@ func (p *Poller) Start(ch <-chan struct{}) error {
 	}
 	p.log.Info("Starting application poller")
 
-	ctx := context.Background()
+	ctx, cancel := context.WithCancel(context.Background())
+
+	// Handle controller-manager signal
+	go func() {
+		for range ch {
+			cancel()
+		}
+	}()
 
 	query := applications.ActivityFeedQuery{}
 	query.SetType(applications.TagScan, applications.TagRun)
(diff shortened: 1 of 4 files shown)


controllers/application_poller_controller.go
```diff
@@ -89,15 +100,21 @@ func (p *Poller) Start(ch <-chan struct{}) error {
...
+				p.handleActivity(ctx, activityItem)
+			}
+		}()
+
+		if err := subscriber.Subscribe(ctx, activityCh); err != nil {
+			// Clean exit/shutdown
+			if errors.Is(err, context.Canceled) {
+				return nil
+			}
+			p.log.Error(err, "Application service connection interrupted, restarting")
```
jgustie, line 115: Do we need to `return nil` if `err == nil`? I don't think it's possible with the current code, but it might be unexpected at a later time.

jgustie (dismissed):

controllers/application_poller_controller.go
```diff
@@ -79,7 +80,14 @@ func (p *Poller) Start(ch <-chan struct{}) error {
...
 	p.log.Info("Starting application poller")
 
-	ctx := context.Background()
+	ctx, cancel := context.WithCancel(context.Background())
+
+	// Handle controller-manager signal
+	go func() {
+		for range ch {
+			cancel()
+		}
```
jgustie, line 89: Just curious as to why this syntax instead of the `<- ch / cancel()`?

jgustie (approved):

controllers/application_poller_controller.go
bradbeam: Real reason, linter yelled at me.
But beyond that it's a bit more concise and simpler; since there's only a single channel we're working with, a for/select construct works but is a bit overkill.

bradbeam: Right after I hit merge, I do wonder if I should have included a break in here. I'd imagine it's not super significant because this should only get triggered on shutdown when the manager notifies us of shutdown, and then we handle the ctx.Done() and return later on.

PR #403, opened 2021-08-25
Quick and dirty hack to avoid loading OpenAPI schema

jgustie:
As advertised, for kinds we know about we can bypass the OpenAPI schema loading. In the current state, we shouldn't fall back to the schema anymore.


jgustie:
Never mind. Doesn't actually help.

PR #414, opened 2021-09-13
Check metric target type if "resourceRequests" is used

jgustie:
Adds an extra check to the linter to catch the case where the `resourceRequests` template function is used but the target is missing or has the wrong type.

Also, the linter now better mimics the actual metric controller in that it defaults the type to trial if the target is missing or has no kind.

### DIFF
--- a/cli/internal/commands/check/experiment.go
+++ b/cli/internal/commands/check/experiment.go
@@ -194,21 +194,7 @@ func (l *linter) Visit(ctx context.Context, obj interface{}) experiment.Visitor 
 		if o.Query == "" {
 			lint.V(vError).Info("Metric query is required")
 		} else {
-			q, _, err := metricQueryDryRun(o)
-			if err != nil {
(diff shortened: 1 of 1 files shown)


cli/internal/commands/check/experiment.go
```diff
@@ -283,12 +269,36 @@ func checkBaseline(lint logr.Logger, p *optimizev1beta2.Parameter) {
 	}
 }
 
-func metricQueryDryRun(m *optimizev1beta2.Metric) (string, string, error) {
+func checkQuery(lint logr.Logger, m *optimizev1beta2.Metric) {
+	switch m.Type {
```
bradbeam, line 273: Is there a reason we wouldn't consolidate this into a single switch statement?

jgustie: They are looking at different things at different times: the first switch looks at the raw query prior to evaluation (and we may return from it); the second switch looks at the evaluated query (template evaluated against an empty object).

bradbeam (approved):

PR #417, opened 2021-09-22
Use updated links to Optimize documentation

jameshochadel:
Looks like the links in internal/application/documentation.go could also use an update, but some aren't published to the docs site yet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,6 @@ This repository contains the StormForge Optimize Controller and CLI implementati
 
 ## Getting Started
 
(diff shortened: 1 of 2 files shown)


jgustie:
Are there additional dead/incorrect links in [here](https://github.com/thestormforge/optimize-controller/blob/master/internal/application/documentation.go)? I think the `app.yaml` reference was removed for the 2.0 stuff so we might need to just drop those from the comments...

jameshochadel:
Yep, that's what I was referring to — but aren't they describing the new Applications, rather than `app.yaml`? I'm not sure either way. Want to put a commit on this branch with the correct updates?

jgustie (approved):

bradbeam (approved):

PR #418, opened 2021-09-22
Update parameter namer

bradbeam:
This introduces generating parameter names as fully qualified paths, ex `deployment/test/test1/cpu`. 
This (hopefully) should clear up some of the confusion around the discovered parameter names.

### DIFF
--- a/cli/internal/commands/export/export_test.go
+++ b/cli/internal/commands/export/export_test.go
@@ -26,10 +26,14 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"github.com/thestormforge/konjure/pkg/filters"
 	"github.com/thestormforge/optimize-controller/v2/cli/internal/commander"
 	"github.com/thestormforge/optimize-controller/v2/cli/internal/commands/export"
 	"github.com/thestormforge/optimize-go/pkg/config"
+	appsv1 "k8s.io/api/apps/v1"
 	"sigs.k8s.io/kustomize/api/filesys"
+	"sigs.k8s.io/kustomize/kyaml/kio"
+	"sigs.k8s.io/yaml"
 )
 
 func TestPatchExperiment(t *testing.T) {
@@ -106,10 +110,16 @@ func TestPatchExperiment(t *testing.T) {
 			err := cmd.Execute()
 			require.NoError(t, err)
 
+			exp, err := extractDeployment(b.Bytes(), "postgres")
+			require.NoError(t, err)
+
 			cpu := wannabeTrial.TrialAssignments.Assignments[0]
 			mem := wannabeTrial.TrialAssignments.Assignments[1]
-			assert.Contains(t, b.String(), fmt.Sprintf("%s: %sm", cpu.ParameterName, cpu.Value.String()))
-			assert.Contains(t, b.String(), fmt.Sprintf("%s: %sMi", mem.ParameterName, mem.Value.String()))
+			cpuLimits := exp.Spec.Template.Spec.Containers[0].Resources.Limits["cpu"]
+			memLimits := exp.Spec.Template.Spec.Containers[0].Resources.Limits["memory"]
+
+			assert.Equal(t, fmt.Sprintf("%sm", cpu.Value.String()), (&cpuLimits).String())
+			assert.Equal(t, fmt.Sprintf("%sMi", mem.Value.String()), (&memLimits).String())
 		})
 	}
 }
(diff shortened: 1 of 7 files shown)


cli/internal/commands/export/export_test.go
```diff
@@ -182,10 +192,36 @@ func TestPatchApplication(t *testing.T) {
...
+		Outputs: []kio.Writer{kio.ByteWriter{Writer: &deploymentBuf}},
+	}
+	if err := deploymentInput.Execute(); err != nil {
+		return nil, err
+	}
+
+	exp := &appsv1.Deployment{}
+	if err := yaml.Unmarshal(deploymentBuf.Bytes(), exp); err != nil {
+		return nil, err
+	}
```
jgustie, line 224: You can also use an `w := &sfio.ObjectList{}` instead of a `kio.ByteWriter`, then `w.Items[0].Object.(*appsv1.Deployment)` is the result.

internal/experiment/generation/transformer.go
```diff
@@ -289,8 +265,7 @@ func parameterNamer(selected []interface{}) ParameterNamer {
 		// against Go structs, if the template parser encounters a token that is
 		// not a valid Go field name, parsing fails (e.g. "bad character U+002D '-'").
 
-		parameterName := strings.Join(parts, "_")
-		parameterName = strings.ReplaceAll(parameterName, "-", "_")
+		parameterName := strings.Join(parts, "/")
 		parameterName = strings.ToLower(parameterName)
```
jgustie: Highly subjective, but I'd say no global lowercasing is necessary anymore, just lowercase the `meta.Kind` when adding it to `parts` (this impacts the generated names of the environment variables, but now that they can be arbitrary strings, I think I'm OK with that).

internal/experiment/generation/transformer.go
```diff
@@ -289,8 +265,7 @@ func parameterNamer(selected []interface{}) ParameterNamer {
 		// against Go structs, if the template parser encounters a token that is
 		// not a valid Go field name, parsing fails (e.g. "bad character U+002D '-'").
```
jgustie: This comment block should have been removed with #411, you can wipe it out now.

internal/experiment/generation/transformer.go
```diff
@@ -236,47 +237,22 @@ func (p *pnode) TargetRef() *corev1.ObjectReference {
...
-					if _, value, _ := yaml.SplitIndexNameValue(p); value != "" {
-						parts = append(parts, value)
+		// In this instance we have multiple containers in a pod that we're
+		// dealing with, so we need to add container name to the parameter
+		// name to better namespace it
+		for pathIdx, p := range path {
+			if yaml.IsListIndex(p) {
+				if _, value, _ := yaml.SplitIndexNameValue(p); value != "" {
+					if path[pathIdx-1] == "env" {
+						parts = append(parts, "env")
```
jgustie: I don't think we need this, we didn't have it before. As long as the `name` comes in blank we'll just end up with `memory`, `cpu`, `replicas`, `MY_ENV_VAR` all at the same level in the hierarchy (i.e. there is no `env` magic value equivalent in the name hierarchy for container resources or replicas).

internal/experiment/generation/transformer.go
```diff
@@ -135,12 +135,13 @@ func (t *Transformer) Transform(nodes []*yaml.RNode, selected []interface{}) ([]
...
-	if expNode, err := (sfio.ObjectSlice{&exp}).Read(); err != nil {
+	expNode, err := (sfio.ObjectSlice{&exp}).Read()
+	if err != nil {
 		return nil, err
-	} else {
-		result = append(expNode, result...) // Put the experiment at the front
 	}
 
+	result = append(expNode, result...) // Put the experiment at the front
+
```
jgustie: If you're going to break up the scoping-if, at least give the poor comment a promotion to its own line ;)

internal/experiment/generation/transformer.go
bradbeam: When I was thinking through this one, I got a grin like this smug face -- :trollface: 

There's no hard requirement that an environment variable be capitalized, so it's possible someone may use `cpu` or `memory` or `replicas` as the environment variable in addition to wanting to tune requests/resources and we'd wind up with an invalid experiment because of duplicated parameters.

Also I was thinking by injecting `env` there, it'd also make it better for display purposes in the UI .

jgustie: What if they have a container named `env`? :trollface:

bradbeam: 💔

bradbeam: Interestingly enough, I couldn't break it with container name == env, so yay for that.

jgustie: I'm fine to leave the `/env/` in there, but if we do I think we should also be injecting `/resources/` on the `memory` and `cpu` parameters (basically you are adding `path[pathIdx-1]` for something like `pathIdx > 4` instead of `== "env"`). I'm not sure where this leaves us with `replicas` though (I guess since there are not indexed path segments, it just gets a shorter parameter name).

Also, we should watch out for [extended resources](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#extended-resources) even if we aren't fully supporting them in generation quite yet (it's not important that the parameter names be reversible, they just can't be ambiguous).

jgustie (approved):

PR #422, opened 2021-10-21
feat: Use helm instead of kustomize for prometheus

bradbeam:
This transitions the builtin prometheus deployment from using kustomize
to helm. This enables us to expose some of the prometheus configuration
in the setup task.

Previously we would use a task like the following:
```yaml
spec:
  trialTemplate:
    spec:
      setupServiceAccountName: optimize-setup-f19e28
      setupTasks:
      - name: monitoring
        args:
        - prometheus
        - $(MODE)
```

This now becomes
```yaml
spec:
  trialTemplate:
    spec:
      setupServiceAccountName: optimize-setup-f19e28
      setupTasks:
      - name: monitoring
        helmChart: ../prometheus
```

With optionally being able to configure requests/limits

```yaml
spec:
  trialTemplate:
    spec:
      setupServiceAccountName: optimize-setup-f19e28
      setupTasks:
      - name: monitoring
        helmChart: ../prometheus
        helmValues:
        - name: storageVolumeSize
          value: 99999Mi
        - name: kubeStateMetrics.resources.requests.cpu
          value: 2000m
        - name: kubeStateMetrics.resources.requests.memory
          value: 2000Mi
        - name: kubeStateMetrics.resources.limits.cpu
          value: 4000m
        - name: kubeStateMetrics.resources.limits.memory
          value: 4000Mi
        - name: promServer.resources.requests.cpu
          value: 2000m
        - name: promServer.resources.requests.memory
          value: 2000Mi
        - name: promServer.resources.limits.cpu
          value: 4000m
        - name: promServer.resources.limits.memory
          value: 4000Mi
```


config/docker-entrypoint.sh
```diff
@@ -1,20 +1,21 @@
...
-
-    namePrefix="optimize-"
-    if [ -n "$NAMESPACE" ]; then
-      namePrefix="optimize-$NAMESPACE-"
-    fi
-
-    kustomize edit set nameprefix "$namePrefix"
-    waitFn() {
-      kubectl wait --for condition=Available=true --timeout 120s deployment.apps ${namePrefix}prometheus-server
-    }
```
jgustie, line 17: Don't we still need this wait condition?

config/docker-entrypoint.sh
```diff
@@ -1,20 +1,21 @@
-#!/bin/sh
+#!/bin/sh -x
```
jgustie: I'm pretty sure `-x` breaks the `build` mode because the output is no longer pure YAML.

config/prometheus/.helmignore
```diff
@@ -0,0 +1,23 @@
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
```
jgustie: Is this necessary given that it will be inside a container and none of these files will be present?

config/prometheus/templates/deployment.yaml
```diff
@@ -3,25 +3,31 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   labels:
+    app.kubernetes.io/name: optimize
+    app.kubernetes.io/managed-by: stormforge
     app: prometheus
-  name: prometheus-server
+  name: optimize-prometheus-server
```
jgustie: I would have expected `{{ .Release.Name }}-prometheus-server` (or `{{ .Release.Name }}-{{ .Release.Namespace }}-prometheus-server` if we just always set the the release name to "optimize") on all these objects to just take the place of the Kustomize name prefix transformation.

Alternately, I would expect the chart could be simplified by just leaving the names static and still relying on the Kustomize behavior to add the name prefixes rather then trying to re-implement it in Go template helpers.

config/prometheus/templates/deployment.yaml
```diff
@@ -62,14 +63,9 @@ spec:
...
-            cpu: 25m
-            memory: 25M
-          limits:
-            cpu: 100m
-            memory: 50M
+          {{- toYaml .Values.pushGateway.resources | nindent 10 }}
       - name: prometheus-server
-        image: "prom/prometheus:v2.25.0"
+        image: "{{ .Values.promServer.image.repository }}:{{ .Values.promServer.image.tag }}"
         imagePullPolicy: "IfNotPresent"
```
jgustie: It looks like the pull policies are in the values but aren't used, I would either use them or remove them from values.

config/docker-entrypoint.sh
```diff
@@ -1,20 +1,21 @@
...
+    cat <<-EOF >helm.yaml
+		apiVersion: konjure.carbonrelay.com/v1beta1
+		kind: HelmGenerator
+		metadata:
+		  name: prometheus
+		releaseName: prometheus
+		chart: ../prometheus
+		EOF
+
+    export HELM_CONFIG=$(cat helm.yaml | base64 -w0)
```
jgustie: It feels like you are doubling up on writing to the file here, can't you just do something like this?
```
	HELM_CONFIG=$(cat <<-EOF | base64 -w0
		...
		releaseName: optimize
		releaseNamespace: "${NAMESPACE}"
		chart: ../prometheus
	EOF)
```

config/prometheus/templates/_helpers.tpl
```diff
@@ -0,0 +1,62 @@
...
+{{/*
+Create the name of the service account to use
+*/}}
+{{- define "prometheus.serviceAccountName" -}}
+{{- if .Values.serviceAccount.create }}
+{{- default (include "prometheus.fullname" .) .Values.serviceAccount.name }}
+{{- else }}
+{{- default "default" .Values.serviceAccount.name }}
+{{- end }}
+{{- end }}
```
jgustie: I don't think I understand what this whole file is doing.

bradbeam: I was undecided on leaving this helper template file in here. It gets generated by default and contains some common helper functions for charts. I think there's some reuse we can get out of it ( namely around name generation ), but fine dropping it.

config/prometheus/templates/deployment.yaml
bradbeam: The naming is something that we still need to think through. Since we some of the prometheus urls hardcoded in the controller[1], I tried to leave it out of scope for this change and just use something that gets us pretty close to current state ( realistically its just the service name that we care about for those cases ). 

I feel like since these prometheus instances are scoped to the trial, it makes sense to tweak the `HelmGenerator` config and put in `trial.Name` in the release ( not sure why we didn't explore this option when we first introduced the prometheus bits ). In that case it'd make sense to change all the name resources to something like `{{ .Release.Name }}-prometheus-server`. 

https://github.com/thestormforge/optimize-controller/blob/65602f4ac9bd338d5c4822fc1220d028394c9b24/internal/setup/setup.go#L161
https://github.com/thestormforge/optimize-controller/blob/65602f4ac9bd338d5c4822fc1220d028394c9b24/controllers/metric_controller.go#L292

jgustie:
There should probably also be a check added to [`setup.IsPrometheusSetupTask`](https://github.com/thestormforge/optimize-controller/blob/6b3a0c74723b95f87a0a428d87007d5d45a5b672/internal/setup/setup.go#L170) to make sure the Pushgateway URL is set on trial jobs using the customized parameters. Something like `if st.HelmChart == "../promethus"` would probably be enough (but it should go after the image check).

config/docker-entrypoint.sh
bradbeam: Yeah we do. Im not sure how to best implement it yet. 

helm offers up `--wait` when using helm install, but we do `helm template | kubectl create`. So we could potentially add in a `waitFn` in the HELM_CONFIG check to wait for one of the more well known labels -- 

```
    helm.sh/chart: yolo-0.1.0
    app.kubernetes.io/name: yolo
    app.kubernetes.io/instance: RELEASE-NAME
    app.kubernetes.io/version: "1.16.0"
    app.kubernetes.io/managed-by: Helm
```

Unsure if that is too fragile - if a chart doesn't use any of the above labels, what do we wait on?

bradbeam: I guess we could just restrict it to our chart.

jgustie: This is such a private use thing on an area of the code with numbered days; I'd say make the Helm aspect of it as "un-Helm-y" as possible to keep it as close to the previous behavior as possible (it is a ".." release). Really, in a sense, we are just abusing Helm as a Go template evaluator to feed into Kustomize.

config/docker-entrypoint.sh
bradbeam: I had quite a bit of flakiness in getting the heredoc to end when including it like this. 
```
CONFIG=$(cat <<-EOF | base64 -w0
apiVersion: konjure.carbonrelay.com/v1beta1
kind: HelmGenerator
metadata:
  name: prometheus
releaseName: prometheus
chart: ../prometheus
EOF)
cmdsubst pipe heredocd> 
```

jgustie (approved):

PR #427, opened 2021-10-27
Add application/scenario lookups to `stormforge run`

bradbeam:
This adds in integration with the application services so we can generate the experiment with the proper application and scenario labels. 

`stormforge run` output:
```
Please select an application name:

[ ] Unsorted Experiments             (default)
[x] default-nginx                    (01fjch9t8zdvsdc5mfyctzage4)
[ ] voting-app-tester                (01fg298ycx5f40fn87c334esqf)


Please select a scenario name:

[x] simple nginx                     (01fjchd15axdkb7t3cnyg0yn9j)

🎉  Your experiment is ready to run!

Application Name: 01fjch9t8zdvsdc5mfyctzage4
Experiment Name: 01fjch9t8zdvsdc5mfyctzage4-14a0041e
```

Generated experiment:
```
apiVersion: optimize.stormforge.io/v1beta2
kind: Experiment
metadata:
  name: 01fjch9t8zdvsdc5mfyctzage4-14a0041e
  namespace: default
  labels:
    stormforge.io/application: '01fjch9t8zdvsdc5mfyctzage4'
    stormforge.io/objective: 'cost-vs-p95-latency'
    stormforge.io/scenario: '01fjchd15axdkb7t3cnyg0yn9j'
 ```

Note, this does require and updated optimize-go release ( fix in https://github.com/thestormforge/optimize-go/pull/61 )

### DIFF
--- a/cli/internal/commander/commander.go
+++ b/cli/internal/commander/commander.go
@@ -29,6 +29,7 @@ import (
 	optimizev1beta2 "github.com/thestormforge/optimize-controller/v2/api/v1beta2"
 	megahack "github.com/thestormforge/optimize-controller/v2/internal/experiment/generation"
 	"github.com/thestormforge/optimize-go/pkg/api"
+	applications "github.com/thestormforge/optimize-go/pkg/api/applications/v2"
 	experimentsv1alpha1 "github.com/thestormforge/optimize-go/pkg/api/experiments/v1alpha1"
 	"github.com/thestormforge/optimize-go/pkg/config"
 	"golang.org/x/oauth2"
@@ -178,6 +179,38 @@ func SetExperimentsAPI(expAPI *experimentsv1alpha1.API, cfg *config.OptimizeConf
 	}
 
 	*expAPI = experimentsv1alpha1.NewAPI(c)
+	return nil
+}
+
+// SetApplicationsAPI creates a new application API interface from the supplied configuration
+func SetApplicationsAPI(appAPI *applications.API, cfg *config.OptimizeConfig, cmd *cobra.Command) error {
+	ctx := cmd.Context()
+	srv, err := config.CurrentServer(cfg.Reader())
+	if err != nil {
+		return err
+	}
+
+	// NOTE: We should use `srv.Identifier` but technically this version of the configuration
+	// exposes this double counted "/v1/experiments/" endpoint
+	address := strings.TrimSuffix(srv.API.ApplicationsEndpoint, "/v2/applications/")
+
+	// Reuse the OAuth2 base transport for the API calls
+	t, err := cfg.Authorize(ctx, oauth2.NewClient(ctx, nil).Transport)
+	if err != nil {
+		return err
+	}
+
+	c, err := api.NewClient(address, t)
+	if err != nil {
+		return err
+	}
+
+	*appAPI = applications.NewAPI(c)
+
+	if _, err := (*appAPI).CheckEndpoint(ctx); err != nil {
+		return err
+	}
+
 	return nil
 }
 
--- a/cli/internal/commander/commander.go
+++ b/cli/internal/commander/commander.go
@@ -207,10 +207,6 @@ func SetApplicationsAPI(appAPI *applications.API, cfg *config.OptimizeConfig, cm
 
 	*appAPI = applications.NewAPI(c)
 
-	if _, err := (*appAPI).CheckEndpoint(ctx); err != nil {
-		return err
-	}
-
 	return nil
 }
 
--- a/cli/internal/commands/run/internal/messages.go
+++ b/cli/internal/commands/run/internal/messages.go
@@ -59,6 +59,12 @@ type KubernetesNamespacesMsg []string
 // test case names.
 type StormForgeTestCasesMsg []string
 
+// ApplicationMsg contains the application ULID if an existing application is used.
+type ApplicationMsg map[string]string
+
+// ScenarioMsg contains the scenario ULID if an existing scenario is used.
+type ScenarioMsg map[string]string
+
 // ExperimentMsg represents the generated experiment.
 type ExperimentMsg []*yaml.RNode
 
--- a/cli/internal/commands/run/internal/messages.go
+++ b/cli/internal/commands/run/internal/messages.go
@@ -60,10 +60,10 @@ type KubernetesNamespacesMsg []string
 type StormForgeTestCasesMsg []string
 
 // ApplicationMsg contains the application ULID if an existing application is used.
-type ApplicationMsg map[string]string
+type ApplicationMsg []string
 
 // ScenarioMsg contains the scenario ULID if an existing scenario is used.
-type ScenarioMsg map[string]string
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

File: cli/internal/commands/run/run.go [hunk 1]
```diff
@@ -134,6 +140,12 @@ func (o *Options) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		case tea.KeyCtrlC:
 			return o, tea.Quit
 
+		case tea.KeyEnter:
+			if o.generatorModel.ApplicationInput.Focused() {
+				fmt.Println("app input focused, enter hit, listing scenario names")
```
jgustie: Debug statement.

cli/internal/commander/commander.go
```diff
@@ -181,6 +182,38 @@ func SetExperimentsAPI(expAPI *experimentsv1alpha1.API, cfg *config.OptimizeConf
...
+	c, err := api.NewClient(address, t)
+	if err != nil {
+		return err
+	}
+
+	*appAPI = applications.NewAPI(c)
+
+	if _, err := (*appAPI).CheckEndpoint(ctx); err != nil {
+		return err
+	}
```
jgustie: I'd say it isn't worth the explicit check in the CLI: the code path is going to pass through the API soon enough that you aren't buying anything with the added validation.

cli/internal/commands/run/model.go
```diff
@@ -155,6 +160,24 @@ func (m generatorModel) Update(msg tea.Msg) (generatorModel, tea.Cmd) {
 		m.NamespaceInput.Choices = msg
 		m.NamespaceInput.SelectOnly()
 
+	case internal.ApplicationMsg:
+		choices := make([]string, 0)
+		for ulid, name := range msg {
+			choices = append(choices, fmt.Sprintf("%-32s (%s)", name, ulid))
+		}
+		sort.Strings(choices)
```
jgustie: All the formatting and sorting should happen in the async code (i.e. `listApplicationNames`) so the event loop code can just do the assignment.

cli/internal/commands/run/run.go
```diff
@@ -134,6 +140,12 @@ func (o *Options) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		case tea.KeyCtrlC:
 			return o, tea.Quit
 
+		case tea.KeyEnter:
+			if o.generatorModel.ApplicationInput.Focused() {
```
jgustie: I'm pretty sure this should be in the `generatorModel.Update`.

cli/internal/commands/run/internal/messages.go
```diff
@@ -59,6 +59,12 @@ type KubernetesNamespacesMsg []string
 // test case names.
 type StormForgeTestCasesMsg []string
 
+// ApplicationMsg contains the application ULID if an existing application is used.
+type ApplicationMsg map[string]string
+
+// ScenarioMsg contains the scenario ULID if an existing scenario is used.
+type ScenarioMsg map[string]string
+
```
jgustie: These should just be `[]string` since they are going in choice inputs.

cli/internal/commands/run/run.go
```diff
@@ -272,12 +287,25 @@ func (o *Options) ReadApplication(args []string) error {
...
 // applyToApp takes all of the what is on the model and applies it to an application.
 func (m generatorModel) applyToApp(app *optimizeappsv1alpha1.Application) {
+	if m.ApplicationInput.Enabled() {
+		parts := strings.Fields(m.ApplicationInput.Value())
+		app.Name = strings.Trim(parts[len(parts)-1], "()")
+	}
+
+	var scenarioName string
+	if m.ScenarioInput.Enabled() {
+		parts := strings.Fields(m.ApplicationInput.Value())
```
jgustie: `m.ScenarioInput.Value()`

bradbeam: ah yeah, caught that locally and forgot to add :oops:

cli/internal/commands/run/run.go
bradbeam: I thought the same, but opted to put it here since we don't have the application api in the generator model. Any strong preference one way or the other ( easy enough to add the client into generator model )?

cli/internal/commands/run/view.go
```diff
@@ -204,6 +204,24 @@ https://docs.stormforger.com/guides/getting-started/`,
...
 	o.generatorModel.ObjectiveInput.Select(2)
 
+	o.generatorModel.ApplicationInput = out.FormField{
+		Prompt: "Please select an application name:",
+		Instructions: []interface{}{
+			"up/down: select",
+			out.KeyBinding{Key: tea.Key{Type: tea.KeyRunes, Runes: []rune{'x'}}, Desc: "choose"},
+		},
+	}.NewChoiceField(opts...)
+	o.generatorModel.ApplicationInput.Select(0)
```
jgustie: There is nothing to select yet, this is what the `SelectOnly()` is for when you are assigning the choices (dito for scenarios below).

cli/internal/commands/run/model.go
bradbeam: fair enough. at one point I thought it'd make more sense to have the ulid:name mappings available, but given choices are just `[]string`, makes sense.

cli/internal/commands/run/view.go
bradbeam: yeah, noticed that as well in testing -- 
```
panic: runtime error: index out of range [-1]

goroutine 34638 [running]:
github.com/thestormforge/optimize-controller/v2/cli/internal/commands/run.(*Options).listScenarioNames(0xc0000b0000)
	/home/bradbeam/src/github.com/thestormforge/optimize-controller/cli/internal/commands/run/command.go:276 +0x4ad
github.com/charmbracelet/bubbletea.(*Program).Start.func7.1()
	/home/bradbeam/src/github.com/thestormforge/optimize-controller/vendor/github.com/charmbracelet/bubbletea/tea.go:373 +0x29
created by github.com/charmbracelet/bubbletea.(*Program).Start.func7
	/home/bradbeam/src/github.com/thestormforge/optimize-controller/vendor/github.com/charmbracelet/bubbletea/tea.go:372 +0x13d
```

was tempted to add a note here that we don't want to default the selection, but just opted to drop it entirely

cli/internal/commands/run/run.go
jgustie: Got it. This makes sense now.

I guess the "right way" to decouple it would be to have `generatorModel.Update` fire off an "application selected" message which could handled here in `run.go`. Then the generator model stays independent and the main event loop stays clear of handling model specific events.

cli/internal/commands/run/view.go
jgustie: It's the same for the other deferred loading choice boxes, which is a good point: you need to add a `LoadingMessage` to the field definition so it shows a spinner if the choices haven't loaded yet.

jgustie (approved):

PR #433, opened 2021-11-10
Make fix update application labels

jgustie:
This PR adds a migration step to experiments that attempts to match the "application" label to the display name of a known application, if a match is found the label is updated to have the actual application name.


bradbeam (approved):
🔥

PR #434, opened 2021-11-10
feat: Allow ignoring of unschedulable condition via annotation

bradbeam:
This adds in a new annoation, `stormforge.io/skip-unschedulable` to allow us
to ignore unschedulable status.


jgustie:
I'm not sure I like the idea of forcing this on state of the resources we are checking; also I feel like just skipping the check is going to make the readiness timeout unnecessarily large (since it may need to account for auto-scaling of nodes for schedulability in addition to the resource itself actually coming ready once scheduled). I think we should probably consider something like a `clusterAutoscaleDelaySeconds` on the `Trial` CRD so that we can separate the "unschedulable" timeout from the normal readiness timeouts.

PR #438, opened 2021-11-22
refactor: Use MODE environment variable instead of args

bradbeam:
This is mostly whitespace changes; the handling of script args moves to the beginning of the script.
The initial loop to handle args is provided for backwards compatibility with older setup tasks that make use of `prometheus $(MODE)`.

### DIFF
--- a/config/Dockerfile
+++ b/config/Dockerfile
@@ -34,4 +34,3 @@ RUN konjure kustomize init
 
 WORKDIR "/workspace/base"
 ENTRYPOINT ["/workspace/docker-entrypoint.sh"]
-CMD ["build"]
--- a/config/docker-entrypoint.sh
+++ b/config/docker-entrypoint.sh
@@ -1,32 +1,37 @@
 #!/bin/sh
-set -e
+set -ex
 
-case "$1" in
-  prometheus)
-    shift
+waitFn() { :; }
 
-    cat <<-EOF >helm.yaml
-		apiVersion: konjure.carbonrelay.com/v1beta1
-		kind: HelmGenerator
-		metadata:
-		  name: prometheus
-		releaseName: optimize-${NAMESPACE}-prometheus
-		releaseNamespace: ${NAMESPACE}
-		chart: ../prometheus
-		EOF
+while [ "$#" != "0" ] ; do
+  case "$1" in
+    prometheus)
+      shift
 
-    export HELM_CONFIG=$(cat helm.yaml | base64 -w0)
+      export HELM_CONFIG="$(cat <<-EOF | base64
+			apiVersion: konjure.carbonrelay.com/v1beta1
+			kind: HelmGenerator
+			metadata:
+			  name: prometheus
+			releaseName: optimize-${NAMESPACE}-prometheus
+			releaseNamespace: ${NAMESPACE}
+			chart: ../prometheus
+			EOF
+      )"
+    ;;
+    build)
+      shift
+      export MODE=build
+    ;;
+    *)
+      shift
+    ;;
+  esac
+done
 
-    waitFn() {
-      # Wait on {{ releaseName }}-server
-      kubectl wait --for condition=Available=true --namespace "${NAMESPACE}" --timeout 5m deployment.apps "optimize-${NAMESPACE}-prometheus-server"
-    }
-  ;;
-  *)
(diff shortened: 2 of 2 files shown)


config/docker-entrypoint.sh
```diff
@@ -1,32 +1,38 @@
...
+			EOF
+
+      export HELM_CONFIG=$(cat helm.yaml | base64 -w0)
+    ;;
+    build)
+      shift
+      export MODE=build
+    ;;
+    *)
+      shift
```
jgustie, line 27: Are you also going to update the setup task job creation code?

jgustie (dismissed):

config/Dockerfile
```diff
@@ -17,7 +17,7 @@ ENV HELM_URL="https://get.helm.sh/helm-${HELM_VERSION}-linux-amd64.tar.gz" \
     KUSTOMIZE_URL="https://github.com/kubernetes-sigs/kustomize/releases/download/kustomize%2F${KUSTOMIZE_VERSION}/kustomize_${KUSTOMIZE_VERSION}_linux_amd64.tar.gz" \
     KONJURE_URL="https://github.com/carbonrelay/konjure/releases/download/${KONJURE_VERSION}/konjure-linux-amd64.tar.gz"
 
-RUN apk --no-cache add curl && \
+RUN apk --no-cache add curl coreutils && \
```
jgustie: What is coreutils for?

config/docker-entrypoint.sh
```diff
@@ -1,32 +1,38 @@
...
+			apiVersion: konjure.carbonrelay.com/v1beta1
+			kind: HelmGenerator
+			metadata:
+			  name: prometheus
+			releaseName: optimize-${NAMESPACE}-prometheus
+			releaseNamespace: ${NAMESPACE}
+			chart: ../prometheus
+			EOF
+
+      export HELM_CONFIG=$(cat helm.yaml | base64 -w0)
```
jgustie: I don't think I noticed this before, but `-w0` isn't something you can do with the Busy Box `base64`.

bradbeam: Yeah, that's the reason for installing coreutils. I had a similar confused look because I hadn't noticed it either in testing. I put some details in the commit message for https://github.com/thestormforge/optimize-controller/pull/438/commits/80da0f8fe0616d353212851f6de7e2a26e0dd8ed

jgustie: I'd think with quotes you'd be fine, even in ash.
```
export HELM_CONFIG="$(base64 helm.yaml)"
```

jgustie: Or keep the pipe into `base64` and eliminate the file system write
```
export HELM_CONFIG="$(cat <<-EOF | base64
		apiVersion...
		...
		chart: ../prometheus
		EOF
)"
```

config/docker-entrypoint.sh
```diff
@@ -1,32 +1,37 @@
...
+			  name: prometheus
+			releaseName: optimize-${NAMESPACE}-prometheus
+			releaseNamespace: ${NAMESPACE}
+			chart: ../prometheus
+			EOF
+      )"
+    ;;
+    build)
+      shift
+      export MODE=build
```
jgustie, line 24: This case isn't necessary since you dropped the `CMD` in the Dockerfile.

config/docker-entrypoint.sh
```diff
@@ -80,44 +94,37 @@ if [ -n "$TRIAL" ]; then
...
+    waitFn () {
+        if [ -n "$TRIAL" ] && [ -n "$NAMESPACE" ] ; then
+            kubectl wait pods --for=delete --namespace "$NAMESPACE" --selector "stormforge.io/trial=$TRIAL,stormforge.io/trial-role=trialResource" --timeout=5m
+        fi
+    }
+
+    ;;
+  build)
+    handle () { cat ; }
+    waitFn () { :; }
```
jgustie, line 120: This `waitFn` is redundant with the new declaration on line 4. I wonder if there should be a comment on the delete and Prom `waitFn` that the no-op is being redefined?

Another option would be to change the last line to `[[ $(type -t waitFn) == function ]] && waitFn` and skip the no-op declaration, not sure which is more clear.

PR #444, opened 2021-12-14
Allow setting labels and environment variables in setup tasks

bradbeam:
This primarily enables setting labels and environment variables in setup tasks.

The reason for this change is to enable things like proxy settings to be configured on setup tasks as well as application specific labels needed to work with policy engines to allow the pod to be scheduled.

There is also a minor change included with the prometheus chart to allow setting labels as well ( same reasoning as above ) as well as the ability to slightly customize the scrape configs by disabling specific scrapers ( in this case we only want to scrape the push gateway ).

### DIFF
--- a/config/prometheus/templates/deployment.yaml
+++ b/config/prometheus/templates/deployment.yaml
@@ -8,6 +8,9 @@ metadata:
     app.kubernetes.io/instance: {{ .Release.Name }}
     app.kubernetes.io/managed-by: Helm
     app: prometheus
+    {{- with .Values.commonLabels }}
+    {{- toYaml . | nindent 4 }}
+    {{- end }}
   name: {{ .Release.Name }}-server
 spec:
   selector:
@@ -17,6 +20,9 @@ spec:
       app.kubernetes.io/name: optimize-prometheus
       app.kubernetes.io/instance: {{ .Release.Name }}
       app.kubernetes.io/managed-by: Helm
+      {{- with .Values.commonLabels }}
+      {{- toYaml . | nindent 6 }}
+      {{- end }}
   replicas: 1
   template:
     metadata:
@@ -26,6 +32,9 @@ spec:
         app.kubernetes.io/name: optimize-prometheus
         app.kubernetes.io/instance: {{ .Release.Name }}
         app.kubernetes.io/managed-by: Helm
+        {{- with .Values.commonLabels }}
+        {{- toYaml . | nindent 8 }}
+        {{- end }}
     spec:
       serviceAccountName: {{ .Release.Name }}-server
       containers:
--- a/internal/setup/job.go
+++ b/internal/setup/job.go
@@ -31,11 +31,11 @@ import (
 	"sigs.k8s.io/yaml"
 )
 
-// This is overwritten during builds to point to the actual image
+// This is overwritten during builds to point to the actual image.
 var (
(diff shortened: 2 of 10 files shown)


internal/setup/job.go
```diff
@@ -249,3 +244,23 @@ func newHelmGeneratorConfig(task *optimizev1beta2.SetupTask) *helmGeneratorConfi
...
+		for k, v := range task.Labels {
+			if _, ok := labels[k]; ok {
+				continue
+			}
+
+			labels[k] = v
+		}
+	}
+
+	return labels
```
jgustie: I had to double take at the double loop, `ok ... continue` bit, maybe instead:
```
labels := make(map[string]string, len(task.Labels)+3)
for k, v := range task.Labels {
  labels[k] = v
}
labels[optimizev1beta2.LabelExperiment] = expName
labels[optimizev1beta2.LabelTrial] = trialName
labels[optimizev1beta2.LabelTrialRole] = "trialSetup"
return labels
```

internal/setup/job.go
```diff
@@ -88,12 +81,14 @@ func NewJob(t *optimizev1beta2.Trial, mode string) (*batchv1.Job, error) {
...
 			Args:  task.Args,
-			Env: []corev1.EnvVar{
+			Env: append([]corev1.EnvVar{
 				{Name: "NAMESPACE", Value: t.Namespace},
 				{Name: "NAME", Value: task.Name},
 				{Name: "TRIAL", Value: t.Name},
 				{Name: "MODE", Value: mode},
 			},
+				task.Env...,
+			),
```
jgustie: I'm not 100% sure, but I think "last def wins", in which case I think we the explicit values at the end of the list.

internal/setup/setup.go
```diff
@@ -27,18 +27,18 @@ import (
 )
 
 const (
-	// ModeCreate is the primary argument to the setup tools container when the task is creating objects
+	// ModeCreate is the primary argument to the setup tools container when the task is creating objects.
```
jgustie, line 30: This is going to make the linter so happy.

config/prometheus/templates/deployment.yaml
```diff
@@ -8,6 +8,9 @@ metadata:
     app.kubernetes.io/instance: {{ .Release.Name }}
     app.kubernetes.io/managed-by: Helm
     app: prometheus
+    {{- with .Values.commonLabels }}
+    {{- toYaml . | nindent 4 }}
+    {{- end }}
```
jgustie, line 13: Am I correct that `.Values.commonLabels` is totally separate from the setup task labels?

internal/setup/job.go
bradbeam: Went back and forth with which way to go on that. Fine with changing it to the above.

internal/setup/setup.go
bradbeam: > This is going to make the linter so happy.

This was done by the linter to make itself happy :D

config/prometheus/templates/deployment.yaml
bradbeam: Correct. There might be some clever bits we could do in the entry point to try to make it happen automatically, but not sure it was worth the effort / complexity.

jgustie (approved):

PR #445, opened 2021-12-15
feat: Add support for windows cli

bradbeam:
I think there is still some work to do in optimize go ( specifically around XDG_HOME and Kubectl binary name) , but `login` and `generate install` work.

Didn't have a Kubernetes cluster on windows to test with :D

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -6,11 +6,17 @@ builds:
   - dir: cli
     env:
       - CGO_ENABLED=0
+    goos:
+      - windows
+      - darwin
+      - linux
     goarch:
       - amd64
       - arm64
     ignore:
     - goos: linux
+      goarch: arm64
+    - goos: windows
       goarch: arm64
     ldflags:
       - '-s -w'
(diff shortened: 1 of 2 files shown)


jgustie:
Not a fan of this because we don't have the capacity to test it, in the past we have said you should just use the Docker image for the CLI on Windows (similar to what we do to [build the Helm chart](https://github.com/thestormforge/helm-charts/blob/091bd3806a16d7b53788ec9a604ebae6fa996553/build.sh#L16)).

I'm glad the `generate install` works with the embedded file system stuff, that would have been a concern.

Hopefully `optimize-go` isn't doing too much with file paths aside from the initial determination of the path, however like you said we probably need a platform specific alternative to `XDG_CONFIG`.

The `run` command output should be tested (hopefully the libraries handle this for us, but needs testing).

Experiment generation isn't fully exposed anymore, however in a few places I believe it is dependent on the "path" annotation formatting. I think we are good with using `filepath.` with those values, but again, it isn't tested.

Actions should be uploading built artifacts (e.g. [here](https://github.com/thestormforge/optimize-controller/blob/c9c7874ea06d676303fc02777da878e5ccb070b1/.github/workflows/pull_request.yaml#L107); it's worth noting `actions/upload-artifact` might support better syntax for uploading multiple artifacts since those Actions were first written.

PR #447, opened 2021-12-17
Send an explicit image tag to the Helm chart generator

jgustie:
Apparently there is a time delay where `latest` isn't returning the correct image (the last Helm chart build needed to be run manually).


jgustie:
Actually, this wasn't the problem, the problems was I got too cute with the `git commit -F <(./build.sh)` and commit checked the worktree before `build.sh` had a chance to make changes.

bradbeam (approved):

PR #449, opened 2022-01-05
fix: Better handling for application namespace

bradbeam:
This cleans up / corrects the handling of the application namespace selection.


bradbeam:
This is just wrong. I mistook app.Resources to be the things matched by parameters. I mean it's kind of doing the same thing ( with the addition of doing the client list ), but going about it all wrong.

PR #450, opened 2022-01-08
v1beta1 is deprecated in k8s 1.22+

liko9:
`Warning: apiextensions.k8s.io/v1beta1 CustomResourceDefinition is deprecated in v1.16+, unavailable in v1.22+; use apiextensions.k8s.io/v1 CustomResourceDefinition
`

### DIFF
--- a/config/crd/bases/optimize.stormforge.io_experiments.yaml
+++ b/config/crd/bases/optimize.stormforge.io_experiments.yaml
@@ -1,4 +1,4 @@
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:


jgustie:
In general it is not valid to simply change the version number of the resources as there may be structural changes required to the document (e.g. the proposed document does not have a validation schema since the field moved between v1 and v1beta1).

Since this project was scaffolded with KubeBuilder, this change would need to be done using Controller Tools, in particular
these manifests are automatically generated using the `controller-gen` tool. To generate a proper v1 CRD, there would need to be a change in the `Makefile` to pick up a newer version `controller-gen` (which would have a ripple effect on the `controller-runtime` version in use).

All that being said, we still are still offering support for Kubernetes 1.14 and 1.15, neither of which handle a v1 CRD.

liko9:
That's fair and reasonable, sorry for my misunderstanding.  If these manifests are automatically generated, should they be in source control as generated files, or removed in favor of being automatically generated?

Where are the supported versions of Kubernetes documented?  I'm trying to get ahead of the release of 1.22 as some people follow the latest version very closely.

jgustie:
You can see the currently supported versions enumerated in the matrix of the [installation test job](https://github.com/thestormforge/optimize-controller/blob/cf1896ba040570293ea0f097cd226c23b4f5ff9d/.github/workflows/kubematrix.yaml#L8); engineering tracks changes to the support matrix internally.

The committing of generated resources is an artifact of using Kubebuilder, historically it allowed the manifests to be referenced directly in SCM (e.g. `kustomize build github.com/thestormforge/optimize-controller/config/default` though we now provide this functionality through our CLI).

PR #452, opened 2022-02-02
Login force environment

jgustie:
This makes `stormforge login --env x --force` behave more predictably by explicitly overwriting the server and issuer values. This requires a horrible hack to get at the default values, but this is just us getting the far edge of the usefulness for this config code.

### DIFF
--- a/cli/internal/commands/login/login.go
+++ b/cli/internal/commands/login/login.go
@@ -149,6 +149,20 @@ func (o *Options) complete() error {
 		}
 	}
 
+	// Allow `--force --env x` to overwrite existing configuration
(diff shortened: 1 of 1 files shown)


cli/internal/commands/login/login.go
```diff
@@ -149,6 +149,20 @@ func (o *Options) complete() error {
 		}
 	}
 
+	// Allow `--force --env x` to overwrite existing configuration
+	if o.Force && o.Environment != "" {
+		// Avert your gaze.
```
bradbeam, line 154: 👀

bradbeam (approved):

PR #453, opened 2022-02-03
Generate secret should emit extraEnvVars and image pull secrets

jgustie:
This adds any additional environment variables from the configuration to the Helm values.yaml formatted output.

We can consider making the URLs configurable, the original idea was that the Helm chart served a specific production-only audience and so making it configurable didn't make sense.

### DIFF
--- a/cli/internal/commands/authorize_cluster/generator.go
+++ b/cli/internal/commands/authorize_cluster/generator.go
@@ -263,6 +263,24 @@ func printHelmValues(obj interface{}, w io.Writer) error {
 		},
 	}
 
+	// Remove values we already included
+	delete(secret.Data, "STORMFORGE_AUTHORIZATION_CLIENT_ID")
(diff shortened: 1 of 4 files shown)


cli/internal/commands/authorize_cluster/generator.go
```diff
@@ -263,6 +263,24 @@ func printHelmValues(obj interface{}, w io.Writer) error {
 		},
 	}
 
+	// Remove values we already included
+	delete(secret.Data, "STORMFORGE_AUTHORIZATION_CLIENT_ID")
+	delete(secret.Data, "STORMFORGE_AUTHORIZATION_CLIENT_SECRET")
+
+	// The Helm chart hard codes these to the production default values
```
bradbeam: Could we put this behind a (hidden?) flag? Thinking more about internal use case

bradbeam: err I guess my comment is more around including it in `vals` above

bradbeam (approved):

PR #460, opened 2022-05-05
Stop saving the client ID in the configuration

jgustie:
This PR stops writing the client registration URI in the configuration file. The net effect of this is that every call to `gen secret` will produce a unique client ID/secret.

A consequence of this is that the number of clients will grow more quickly so customers may need to revoke unused clients from the "Clusters" tab in the UI.

### DIFF
--- a/cli/internal/commands/authorize_cluster/generator.go
+++ b/cli/internal/commands/authorize_cluster/generator.go
@@ -17,6 +17,8 @@ limitations under the License.
 package authorize_cluster
 
 import (
+	"bufio"
+	"bytes"
 	"context"
 	"encoding/base64"
 	"encoding/json"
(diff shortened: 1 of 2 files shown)


cli/internal/commands/authorize_cluster/generator.go
```diff
@@ -272,6 +272,41 @@ func (o *GeneratorOptions) clientInfo(ctx context.Context, ctrl *config.Controll
 	return o.Config.RegisterClient(ctx, client)
 }
 
+// clusterClientInformation reads the registered client from the cluster, allowing it to be re-used.
+func (o *GeneratorOptions) clusterClientInformation(ctx context.Context, ctrl *config.Controller) *registration.ClientInformationResponse {
+	cmd, err := o.Config.Kubectl(ctx, "get", "secret", "--namespace", ctrl.Namespace, o.Name,
+		"--output", "go-template", "--template",
+		"{{ .data.STORMFORGE_AUTHORIZATION_CLIENT_ID }}/{{.data.STORMFORGE_AUTHORIZATION_CLIENT_SECRET }}")
```
bradbeam: unsure if we want to get cute or not, but we could potentially handle the decoding inside the template --
```
kubectl get secret ...   -o go-template --template '
{{- $id := .data.STORMFORGE_AUTHORIZATION_CLIENT_ID -}}
{{- $secret := .data.STORMFORGE_AUTHORIZATION_CLIENT_SECRET -}}
{{- if and $id $secret }}{{ $id | base64decode }}/{{ $secret | base64decode }}{{ end -}}
'
```

more just a fyi, not requesting change

bradbeam (dismissed):

bradbeam (approved):

PR #461, opened 2022-05-12
Create missing application and scenario

jgustie:
This requires a new optimize-go release.


controllers/server_controller.go
```diff
@@ -266,6 +286,57 @@ func (r *ServerReconciler) createExperiment(ctx context.Context, log logr.Logger
...
+
+	// Check to see if the application exists
+	app, err := r.ApplicationsAPI.GetApplicationByName(ctx, appName)
+	if controller.IgnoreNotFound(err) != nil {
+		return err
+	}
+
+	// A non-nil error here means it was not found, try creating it
+	if err != nil {
+		if _, err := r.ApplicationsAPI.UpsertApplicationByName(ctx, appName, applications.Application{}); err != nil {
```
bradbeam, line 305: should we add in namespace name here from experiment?

bradbeam (approved):

controllers/server_controller.go
jgustie: I don't think the experiment's name doesn't necessarily correspond to the application resources: even if they did I'm not sure we want to make the association via this path to avoid confusion with applications coming from other workflows.

PR #467, opened 2022-05-23
Applications run

jgustie:
This PR makes some updates the `run` command:
1. Improved support for `ChoiceInput` when `Editable == true`
2. Allow `ChoiceInput` choices to be `"value\tName"` where the console shows "Name" but the code sees "value"
3. Moved the application name to top of the field list and made it an editable `ChoiceInput`
4. Moved the scenario name above the scenario type input and changed it from `ChoiceInput` to a `TextInput`

### DIFF
--- a/cli/internal/commands/run/form/validation.go
+++ b/cli/internal/commands/run/form/validation.go
@@ -171,3 +171,7 @@ func (v *Name) ValidateTextField(value string) tea.Msg {
 
 	return ValidationMsg("")
 }
+
+func (v *Name) ValidateChoiceField(value string) tea.Msg {
+	return v.ValidateTextField(value)
+}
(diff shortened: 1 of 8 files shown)


cli/internal/commands/run/form/validation.go
```diff
@@ -152,3 +152,26 @@ func (v *ContainerImage) ValidateTextField(value string) tea.Msg {
...
+
+func (v *Name) ValidateTextField(value string) tea.Msg {
+	if value == "" {
+		return ValidationMsg(v.Required)
+	}
+
+	if !nameRegexp.MatchString(value) {
+		return ValidationMsg(v.Valid)
+	}
+
```
bradbeam, line 171: Should we add in a length check (63 char) here as well?

bradbeam (dismissed):

cli/internal/commands/run/form/validation.go
jgustie: It's in the regexp (and I fixed the length, it was validating to a max of 64 characters).

bradbeam (approved):

PR #468, opened 2022-05-27
Add missing requests error rate to default

jgustie:
The `enforceErrorRate` function that adds the default error rate objective was omitting the type of error rate that should be optimized for (i.e. "requests"). Since Perftest only supports error rate objectives for requests, the Perftest objective code didn't think it could generate the required metric.


txpdev (dismissed):
LGTM

PR #474, opened 2022-06-03
Prometheus support for explicit kubeconfig secret

jgustie:
This PR has two changes:
1. If the scrapes are disabled for KSM or PG, then we won't run the containers that aren't being scraped
2. There is a new `promServer.kubeconfigSecret` value that can be used to override what KSM and Prometheus `kubernetes_sd_configs` use as a kubeconfig

### DIFF
--- a/config/prometheus/templates/deployment.yaml
+++ b/config/prometheus/templates/deployment.yaml
@@ -38,6 +38,7 @@ spec:
     spec:
       serviceAccountName: {{ .Release.Name }}-server
       containers:
+      {{- if .Values.promServer.scrapes.kubeStateMetrics }}
       - name: kube-state-metrics
         args:
         - --collectors=pods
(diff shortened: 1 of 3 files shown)


config/prometheus/templates/deployment.yaml
```diff
@@ -59,10 +60,17 @@ spec:
           timeoutSeconds: 5
         resources:
           {{- toYaml .Values.kubeStateMetrics.resources | nindent 10 }}
+        {{- if .Values.promServer.kubeconfigSecret }}
+        volumeMounts:
+        - name: kubeconfig-volume
+          mountPath: /.kube
```
bradbeam, line 66: Unsure if root's home here is `/` or `/root`
may also want to add in `automountServiceAccountToken: false` on the pod spec if we aren't going to use the built in service account

bradbeam (approved):

PR #477, opened 2022-06-07
Make sure application labels are valid

jgustie:
This is a quick and dirty validation to make sure application/scenario labels are valid on an experiment. While this might have failed further on, putting the validation here ensures we don't actually create corrupt objects in the API.

This validation will be cleaned up in. subsequent release.

### DIFF
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"math"
 	"path"
+	"regexp"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


internal/server/server.go
```diff
@@ -38,13 +39,26 @@ const (
 	Finalizer = "serverFinalizer.stormforge.io"
 )
 
+var (
+	// nameRegexp is used to validate the experiment labels that are required to be names.
+	nameRegexp = regexp.MustCompile(`^[a-z\d](?:[-a-z\d]{0,61}[a-z\d])?$`)
```
bradbeam, line 44: should we have this in optimize-go? 
I don't think it's a blocker for this pr.

bradbeam (approved):

internal/server/server.go
jgustie: [Yes](https://github.com/thestormforge/optimize-go/commit/6ff58273bd24d70178fb9cee874819b5f1d19158) ;)

PR #479, opened 2022-06-08
Upgrade default versions for the built-in Prometheus

jgustie:


config/prometheus/values.yaml
```diff
@@ -61,7 +61,7 @@ configReload:
   image:
     repository: jimmidyson/configmap-reload
     pullPolicy: IfNotPresent
-    tag: v0.5.0
+    tag: v0.7.1
```
jgustie, line 64: This was really just a pair of dependency bumps and a rename of the `master` branch to `main`.

bradbeam: Just going to comment here to get it out of my head (nothing to do for this pr), but we might want to look at shifting over to using 

https://quay.io/repository/prometheus-operator/prometheus-config-reloader?tab=tags&tag=latest // 
https://github.com/prometheus-operator/prometheus-operator/blob/main/cmd/prometheus-config-reloader/main.go

bradbeam (approved):

PR #482, opened 2022-06-09
Expose constant parameters as assignments

jgustie:
For a long time, we have had the ability to set a parameter min==max to have it omitted from the experiment definition we send to the server. However, this did not work for categorical parameters (i.e. a parameter with a single value). Also, because the parameter was withheld from the server definition, it was not included in the assignments that come back and therefore could not be used in patches.

This PR fixes the issue with omitting single value categorical parameters and adds the constant value into the trial assignments so it can be used in patches.

### DIFF
--- a/internal/experiment/trial_template.go
+++ b/internal/experiment/trial_template.go
@@ -20,6 +20,7 @@ import (
 	optimizev1beta2 "github.com/thestormforge/optimize-controller/v2/api/v1beta2"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
 // PopulateTrialFromTemplate creates a new trial for an experiment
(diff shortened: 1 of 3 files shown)


internal/experiment/trial_template.go
```diff
@@ -59,4 +60,31 @@ func PopulateTrialFromTemplate(exp *optimizev1beta2.Experiment, t *optimizev1bet
...
+}
+
+// ParameterConstant returns the value of the supplied parameter if it is constant.
+// A constant parameter has a domain which includes exactly 1 value (i.e. min==max
+// for numeric, or len(values)==1 for categorical); these parameters are not processed
+// on the server.
+func ParameterConstant(p optimizev1beta2.Parameter) *intstr.IntOrString {
+	switch {
+	case p.Min != p.Max || len(p.Values) > 1:
+		return nil
```
bradbeam: nit, I'd switch around this case with the default so we highlight the actual cases we care about, 

```go
switch {
case p.Min == p.Max && len(p.Values) == 0:
		v := intstr.FromInt(int(p.Max))
 		return &v
case len(p.Values) == 1:
		v := intstr.FromString(p.Values[0])
 		return &v
default:
		return nil
}
```

bradbeam (approved):

PR #495, opened 2022-09-14
Fix `set config` for controllers with dot names

reidmv:
Previously if a controller had a dot character "." in the name, it wasn't possible to set configs on that controller.

This commit fixes the issue by using optimize-go 0.0.28, which is assumed to support escaping dot characters in path element names with a backslash. E.g. "one\.two" is a single path element.

NOTE: This PR depends on thestormforge/optimize-go#85, and a new release of optimize-go. It is in draft form until that condition is met.


jgustie:
This has been superseded by the v3.0.0 CLI release.

PR #498, opened 2022-09-29
Point check version at downloads site

jgustie:
This updates the `check version` command to look at `downloads.stormforge.io` instead of GitHub for the latest release so we recognize when v3 is available.

### DIFF
--- a/cli/internal/commands/check/version.go
+++ b/cli/internal/commands/check/version.go
@@ -18,22 +18,17 @@ package check
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
-	"net/http"
+	"net/url"
 	"path"
 	"runtime"
-	"strings"
 
+	"github.com/mmcdole/gofeed"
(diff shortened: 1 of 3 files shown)


cli/internal/commands/check/version.go
```diff
@@ -56,105 +51,28 @@ func NewVersionCommand(o *VersionOptions) *cobra.Command {
...
-		if r.Assets[i].Name == name {
-			return &r.Assets[i]
+		downloadURL, err := url.Parse(item.Link)
+		if err != nil {
+			return err
 		}
-	}
-	return nil
-}
+		downloadURL.Path = path.Join(downloadURL.Path, fmt.Sprintf("stormforge_%s_%s_%s.tar.gz", item.Title, runtime.GOOS, runtime.GOARCH))
```
bradbeam, line 69: i wonder if we may want to expose these ( OS / Arch ) as flags in the future

bradbeam (approved):

